Issue #33, opened 2026-05-24
certs error in docker compose setup


shokri-navid:
I tried floci-az by complying the instruction in [this page](https://floci.io/floci-az/getting-started/quick-start/). and run the service using this command: 

```bash 
    docker compose -f compose.yaml up  floci-az
```

But it raises this error in startup step and stop immediately:

2026-05-24 21:55:11,850 ERROR [io.qua.run.Application] Failed to start application: java.lang.RuntimeException: Failed to start quarkus
floci-az  | 	at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
floci-az  | 	at io.quarkus.runtime.Application.start(Application.java:112)
floci-az  | 	at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:127)
floci-az  | 	at io.quarkus.runtime.Quarkus.run(Quarkus.java:79)
floci-az  | 	at io.quarkus.runtime.Quarkus.run(Quarkus.java:50)
floci-az  | 	at io.quarkus.runtime.Quarkus.run(Quarkus.java:143)
floci-az  | 	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
floci-az  | Caused by: java.nio.file.NoSuchFileException: certs/floci-az.p12
floci-az  | 	at java.base@25.0.3/sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:213)
floci-az  | 	at java.base@25.0.3/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:244)
floci-az  | 	at java.base@25.0.3/java.nio.file.Files.newByteChannel(Files.java:357)
floci-az  | 	at java.base@25.0.3/java.nio.file.Files.newByteChannel(Files.java:399)
floci-az  | 	at java.base@25.0.3/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:371)
floci-az  | 	at java.base@25.0.3/java.nio.file.Files.newInputStream(Files.java:154)
floci-az  | 	at io.quarkus.vertx.http.runtime.options.HttpServerOptionsUtils.getFileContent(HttpServerOptionsUtils.java:447)
floci-az  | 	at io.quarkus.vertx.http.runtime.options.TlsUtils.createKeyStoreOptions(TlsUtils.java:161)
[...]

PR #36, opened 2026-05-25
feat(tls): dynamic self-signed certificate generation at runtime

hectorvent:
## Summary

- Dynamic TLS cert generation: Extracts cert generation into a CertificateGenerator class (matching aws-local structure) backed by BouncyCastle. TlsConfigSource delegates to it; certs are generated at startup, persisted under data/tls/, and regenerated when hostname config changes.
- TLS proxy: TlsProxyServer sniffs the first byte of each connection (0x16 = TLS) and routes to the HTTPS or HTTP internal backend — both protocols on the same public port 4577.
- /_floci/tls-cert endpoint: Added to HealthController so compat tests can fetch the active cert PEM and install it in their truststore dynamically.
- Removed static certs: src/main/resources/certs/ and compatibility-tests/sdk-test-java/src/test/resources/floci-az.p12 deleted — no longer needed.
- GraalVM native support: Added --initialize-at-run-time flags for BouncyCastle classes and CertificateGenerator; BouncyCastleInitializer CDI bean ensures the BC provider is registered before use.
- Event Hubs mocked mode: NamespaceState gains a mocked field; management API returns "mocked":true when no Artemis broker is running, allowing compat tests to skip AMQP tests gracefully.
- Cosmos compat fix: EmulatorConfig.buildCosmosClient() now fetches and installs the emulator cert before building the client; Netty forced to JDK SSL via -Dio.netty.handler.ssl.noOpenSsl=true so javax.net.ssl.trustStore is respected.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Checklist


hectorvent:
fixes #33

Issue #62, opened 2026-06-13
Add Azure Container Apps support


hectorvent:
## Title

Add Azure Container Apps support

## Description

Implement Azure Container Apps support inside floci-az.

The purpose is Docker-backed apps with revisions, ingress, and scale rules — the Azure counterpart of AWS App Runner / ECS and GCP Cloud Run. This sits on top of the ACI + ACR groundwork.

ARM providers: `Microsoft.App/managedEnvironments` + `Microsoft.App/containerApps`.

Suggested capabilities:

* Managed environment CRUD
* Container App CRUD (template, containers, env, secrets)
* Revisions (single / multiple revision mode)
* Ingress (external/internal, target port, FQDN) and HTTP routing to the running container
* Scale rules (min/max replicas) — best-effort locally

floci-az notes: build on the ACI container lifecycle and pull images via the `acr` registry sidecar; high complexity. Depends on the ACI issue.

## Official Documentation

* https://learn.microsoft.com/en-us/azure/container-apps/
* https://learn.microsoft.com/en-us/rest/api/containerapps/

## Azure SDK for Java

* Management: `com.azure.resourcemanager:azure-resourcemanager-appcontainers` — https://learn.microsoft.com/en-us/java/api/overview/azure/resourcemanager-appcontainers-readme

## Azure REST API Specs

* https://github.com/Azure/azure-rest-api-specs/tree/main/specification/app

yaswanthkumarch:
@hectorvent I’d like to work on this issue. Please assign it to me—I’ll start implementation right away.

PR #76, opened 2026-06-19
feat(containerapps): add Azure Container Apps MVP (CRUD, ingress, scaling, revisions)

yaswanthkumarch:
## Summary
This PR adds **scaffolding** for Azure Container Apps support in *floci‑az*.  
It introduces a `ContainerAppService` bean that implements `AzureServiceHandler` and currently returns a **NotImplemented** response for all Container App operations.

### What was changed
* Removed the unused `ContainerAppsClient` injection from `AzureServiceRegistry`.
* Deleted the unused Maven dependency `azure‑resourcemanager‑appcontainers:1.0.0`.
* Fixed indentation/formatting in `AzureServiceRegistry`.
* Updated comments to state that the implementation is **only a skeleton** (no CRUD, ingress, scaling, or revision logic yet).


hectorvent:
Hey, thanks for taking this on! I took a close look and I think the description and the code have drifted apart, so I want to flag it before we go further.

Right now the code is essentially scaffolding. `ContainerAppService.handle()` returns `NotImplemented` for everything (its own comments say "placeholder"), and there are no models, CRUD, ingress, scaling, or revisions yet. There also aren't any tests in the diff, so the "full unit test coverage" line in the description doesn't match what's here. Could we reword the summary to reflect that this is a first skeleton? That'll save the next reviewer some confusion.

A few things that block it as is:

- The handler can never actually be reached. Container Apps is an ARM resource (`Microsoft.App/containerApps`), so it should be dispatched inside `ArmHandler` by path, the same way `Microsoft.Web/sites` already is. As a `serviceType` handler there's no routing suffix and no `isEnabled()` case, so nothing ever calls it.
- `ContainerAppsClient` is injected into `AzureServiceRegistry` but never used. It's a wrapper around a bean that's already injectable, so I'd just drop it.
- The new `azure-resourcemanager-appcontainers:1.0.0` dependency isn't imported anywhere, and the version is pinned outside the Azure BOM. Probably best to remove it until something uses it.
- Small thing: the new field and assignment in `AzureServiceRegistry` lost their indentation.

If the goal is just to land the scaffolding, happy to approve once the description matches and the dead client and dependency are gone. If you'd rather build it out, the cleaner path is ARM routing through `ArmHandler` on top of the ACI and ACR work, per issue #62. Glad to pair on either direction.

yaswanthkumarch:
@hectorvent Thanks for the detailed feedback.

I've addressed the items you called out:

* Removed the unused `ContainerAppsClient` injection.
* Removed the unused `azure-resourcemanager-appcontainers` dependency.
* Fixed formatting in `AzureServiceRegistry`.
* Updated `ContainerAppService` comments to clearly indicate this is scaffolding only.
* Updated the PR description to align with the current implementation.

This PR should now be treated as Container Apps scaffolding rather than a complete MVP implementation. The actual ARM routing, CRUD operations, ingress, scaling, revisions, and tests will be implemented in follow-up work.

Could you take another look when you have a chance?

hectorvent (approved):
Thanks @yaswanthkumarch

PR #95, opened 2026-06-26
ci: run compatibility tests against native image

electrum:
Builds the native emulator image in compatibility CI and runs the existing compatibility matrix against it on native-architecture test containers. When the emulator runs in Docker, ACR and Redis readiness checks use sidecar DNS names on the shared Docker network.


hectorvent (approved):
Thanks!

PR #96, opened 2026-06-26
ci: run compatibility clients with emulation

electrum:
Runs compatibility client containers as amd64 under QEMU so the matrix works on ARM runners.


electrum:
Folded this change into #95.

PR #98, opened 2026-06-27
fix: make TLS-required guidance discoverable for Terraform/azurerm users (#94)

### LINKED ISSUE #94: [BUG]
Environment:
- Terraform 1.14.3
- AzureRM 4.79.0
- Floci-AZ 0.8.0

Configuration:
ARM_METADATA_HOSTNAME=localhost:4577

Observed:
AzureRM discovers Floci metadata:
"Configuring cloud environment from Metadata Service at localhost:4577"

Then attempts:
https://localhost:4577/metadata/endpoints?api-version=2022-09-01

Floci serves only HTTP and returns:
http: server gave HTTP response to HTTPS client

TLS endpoint:
http://localhost:4577/_floci/tls-cert

returns:
{"error":"TLS not enabled or certificate not available"}

Ali-Shaikh:
Fixes #94.

## Problem

With `ARM_METADATA_HOSTNAME=localhost:4577`, the azurerm provider configures itself by
fetching the cloud metadata **over HTTPS** (`GET https://localhost:4577/metadata/endpoints`).
floci-az serves plain HTTP unless `FLOCI_AZ_TLS_ENABLED=true`, so the TLS handshake fails:

```
Configuring cloud environment from Metadata Service at localhost:4577
http: server gave HTTP response to HTTPS client
```

The handshake fails before any HTTP handler runs, so floci-az can't answer with a friendly
error on that request — and the one diagnostic users reach for, `GET /_floci/tls-cert`,
returned a vague `{"error":"TLS not enabled or certificate not available"}` with no hint that
TLS simply needs enabling, leaving the requirement effectively undiscoverable.

## Fix

TLS is required for the azurerm provider (it controls the scheme for metadata discovery), so
this makes that requirement discoverable:

- **Actionable `/_floci/tls-cert` response.** It's now TLS-state-aware: when TLS is disabled it
  names `FLOCI_AZ_TLS_ENABLED=true`, explains the azurerm metadata-over-HTTPS reason, and links
  the new guide; when enabled-but-not-yet-generated it says to retry. Adds a `tlsEnabled`
  boolean to the body.
- **New `docs/terraform.md` guide** — enabling TLS, trusting the cert, the full provider block
  (`environment="stack"`, `metadata_host`, …), and a troubleshooting section keyed on the exact
  error above. Linked from the README TLS section and added to the mkdocs nav.

## Tests

- Adds a `HealthTest` case asserting the disabled-TLS response carries `tlsEnabled=false` and
  names `FLOCI_AZ_TLS_ENABLED=true`.
- Full suite green (236). No behavioural change beyond the error-body content; the compat
  bootstrap scripts use `curl -sf` on `/_floci/tls-cert` and fall back on non-200, so they are
  unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ services:
 
 The self-signed certificate is generated at startup and cached under `data/tls/`. Fetch it at runtime from `GET http://localhost:4577/_floci/tls-cert` to install it into your truststore — no static cert to bundle or import manually.
 
+> **Terraform / OpenTofu (azurerm) also requires TLS** — the provider discovers the cloud over HTTPS (`GET https://<host>/metadata/endpoints`), so it fails against plain HTTP. See the [Terraform / OpenTofu guide](https://floci.io/floci-az/terraform/).
+
 </details>
 
 ## Features
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -49,7 +49,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~> 3.0"
+      version = ">= 3.0"   # works with the 3.x and 4.x provider lines
     }
   }
 }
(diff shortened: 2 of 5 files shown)


hectorvent (approved):
Hi @Ali-Shaikh, 

Thanks for your contribution. This what makes floci better.

PR #99, opened 2026-06-28
fix: make Azure SQL data plane reachable on a shared Docker network

Ali-Shaikh:
## Summary

Makes the Azure SQL data plane reachable when floci-az runs inside a container
(docker-compose / CI), resolving the same issue fixed for PostgreSQL Flexible Server in #97.

The `azure-sql-edge` sidecar is reachable by other containers via its name on the shared Docker
network, not `localhost:<published-port>`. `SqlConnectionInfo` / `fullyQualifiedDomainName`
hardcoded `localhost`, so applications in a networked deployment could not connect to the
database, and the readiness probe timed out against an unreachable `localhost`.

## Change

Following the existing `RedisCacheManager` pattern: attach the container to the configured
Docker network, and report the container name + container port as the reachable endpoint when
running inside a container (`ContainerDetector`), otherwise `localhost` + the published port.
`/connect`, `fullyQualifiedDomainName`, and the readiness probe use the reachable host.
`SqlServerEntry` gains a non-persisted `host` field.

## Testing

Full test suite passes; `SqlState` tests updated for the new field. The identical PostgreSQL
change in #97 is covered end to end by the native compatibility matrix.

### DIFF
--- a/src/main/java/io/floci/az/services/sql/SqlHandler.java
+++ b/src/main/java/io/floci/az/services/sql/SqlHandler.java
@@ -183,7 +183,7 @@ public class SqlHandler implements AzureServiceHandler {
                 var firewallRules = new java.util.concurrent.ConcurrentHashMap<String, SqlState.SqlFirewallRule>();
                 entry = new SqlState.SqlServerEntry(
                     serverName, sub, rg, location, login, password,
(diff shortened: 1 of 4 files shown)


hectorvent (approved):
Thanks @Ali-Shaikh

PR #104, opened 2026-07-03
feat: add Azure Database for MySQL, MariaDB, and diagnostic settings emulation

guilhermefacanha:
## Summary

- Implements ARM management API stubs for **Azure Database for MySQL (Flexible Server)** and **Azure Database for MariaDB (Single Server)**: Docker container lifecycle, CRUD endpoints (servers, databases, firewall rules, configurations, connection strings, `checkNameAvailability`)
- Extends `MonitorHandler` with a cross-cutting **diagnostic settings extension resource** (`PUT/GET/LIST/DELETE` on `{resourceUri}/providers/microsoft.insights/diagnosticSettings`) covering MySQL, MariaDB, and PostgreSQL, enabling DSF Agentless Gateway to route audit logs to Event Hubs
- Adds `AzureRoutingFilter` to correctly dispatch nested extension resource paths (diagnostic settings) before provider-specific blocks

## Test plan

- [ ] `./mvnw test` — all existing tests pass
- [ ] `MySqlHandlerMockedTest` — server CRUD, database CRUD, firewall rules, configurations, connection strings, checkNameAvailability
- [ ] `MySqlStateTest` — state machine transitions, port allocation, cleanup
- [ ] `MariaDbHandlerMockedTest` — same coverage as MySQL (Single Server API shape)
- [ ] `MariaDbStateTest` — state machine and lifecycle
- [ ] `MonitorHandler` diagnostic settings: PUT, GET, LIST, DELETE for MySQL, MariaDB, and PostgreSQL resources
- [ ] No tests needed for `AzureRoutingFilter` (routing logic is exercised by all handler tests end-to-end)

### DIFF
--- a/src/main/java/io/floci/az/config/EmulatorConfig.java
+++ b/src/main/java/io/floci/az/config/EmulatorConfig.java
@@ -122,6 +122,8 @@ public interface EmulatorConfig {
         EventHubConfig         eventHub();
         SqlServiceConfig       sql();
         PostgresServiceConfig  postgres();
+        MySqlServiceConfig     mysql();
+        MariaDbServiceConfig   mariaDb();
         ServiceBusConfig       serviceBus();
         AksConfig              aks();
         VmConfig               vm();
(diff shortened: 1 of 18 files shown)


hectorvent:
Thanks @guilhermefacanha — really solid work. It follows the postgres pattern closely and the test coverage is great. A few things before this can go in:

1. **The `/connect` endpoints and the `localPort` property aren't part of the real Azure API** — neither the mysql nor the mariadb spec has any connection-string endpoint. We have a hard rule against non-Azure shapes, so please drop them; clients build connection strings from `fullyQualifiedDomainName` + credentials, and the `docker port` is discoverable via the container name.

2. **Routing order bug:** the diagnosticSettings block in `AzureRoutingFilter` sits *after* the `Microsoft.Sql` block, so `.../providers/Microsoft.Sql/servers/x/providers/microsoft.insights/diagnosticSettings/audit` gets swallowed by SqlHandler and 404s — even though `defaultLogCategories()` supports SQL. Move the diag check above all the provider blocks (it's the most specific match).

3. **`defaultLogCategories()` invents enabled log categories when `logs` is omitted.** Terraform's read-back then sees enabled logs it never configured and produces a perpetual diff. Just store what was sent — an empty list stays empty.

4. **`state: "Creating"` isn't in the MySQL ServerState enum** (`Ready|Dropping|Disabled|Starting|Stopping|Stopped|Updating`) — use `Starting`. And if Terraform support matters: azurerm's read hard-fails without these fields, e.g.

   ```java
   props.put("storage", Map.of("storageSizeGB", s.storageSizeGB(),
       "autoGrow", "Disabled", "autoIoScaling", "Enabled", "logOnDisk", "Disabled"));
   props.put("backup", Map.of("backupRetentionDays", 7, "geoRedundantBackup", "Disabled"));
   ```

   (FYI azurerm removed MariaDB entirely, so that one is SDK/CLI-only either way.)

5. **The `QueueServiceHandler` `_visibleAt` seconds→millis change looks unrelated** — could you split it into its own PR with a short note on why? As-is it also misreads previously persisted values (seconds parsed as millis → hidden messages become visible immediately).

Also missing the usual new-service wiring: `application.yml` blocks, `BannerLogger` entries, a docs page + CHANGELOG line, and ideally a compat test (the test-plan boxes are still unchecked 🙂).

Heads-up: a couple of in-flight refactors will conflict here (AdminController reset via `Resettable`, shared ARM helpers in `core/arm`) — happy to help with the rebase once those land.

hectorvent:
Hi @guilhermefacanha — thank you so much for this contribution! The MySQL/MariaDB emulation and the diagnostic-settings routing are solid work, and the thorough test coverage made it easy to build on.

Since the branch had drifted behind main's routing refactor and we haven't heard back in a while (no worries at all — we know how it goes!), we've ported your commits onto the current architecture and will land them in follow-up PRs shortly. Your authorship is preserved on all the ported commits, and we split out your queue visibility-timestamp race fix into its own PR so it gets credited separately in the changelog.

Closing this one in favor of the follow-ups, which will reference this PR. Thanks again for the contribution — you're welcome back anytime! 🙏

PR #107, opened 2026-07-08
fix(core): reset all services via Resettable self-registration

hectorvent:
## Summary

`POST /_admin/reset` hand-listed 18 handlers and silently skipped **apim** and **email**, so their state survived test-isolation resets (apim had no `clearAll` at all; email had one that was never called).

State-holding handlers now implement a new `core/Resettable` interface and `AdminController` iterates CDI `Instance<Resettable>` — the same self-registration pattern `AuthPipeline` uses for `AuthVerifier`. A service registers for reset by existing; forgetting is no longer silently possible. ApiManagement gains its missing `clearAll` (handler delegating to the service, mirroring network/eventgrid).

**Behavior change (intentional):** apim and email state is now actually wiped by `/_admin/reset`.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

No wire-protocol changes — `/_admin/reset` is an emulator-internal endpoint. Each `clearAll()` implementation is unchanged; only the dispatch mechanism moved from a hand-maintained list to CDI discovery.

## Checklist

- [x] `./mvnw test` passes locally (251/251)
- [x] New or updated integration test added (`AdminResetTest`: creates an apim service + an appconfig key, resets, asserts both are gone — the apim half fails on main)

### DIFF
--- a/src/main/java/io/floci/az/core/AdminController.java
+++ b/src/main/java/io/floci/az/core/AdminController.java
@@ -1,51 +1,25 @@
 package io.floci.az.core;
 
-import io.floci.az.services.appconfig.AppConfigHandler;
-import io.floci.az.services.acr.AcrHandler;
-import io.floci.az.services.aks.AksHandler;
-import io.floci.az.services.blob.BlobServiceHandler;
-import io.floci.az.services.cosmos.CosmosHandler;
-import io.floci.az.services.eventgrid.EventGridHandler;
-import io.floci.az.services.eventhub.EventHubHandler;
-import io.floci.az.services.functions.FunctionsServiceHandler;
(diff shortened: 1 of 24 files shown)


hectorvent:
Applied both suggestions in the amended commit:

- `reset()` is now best-effort — each `clearAll()` is wrapped in try/catch, a failing handler is logged with its class name and the loop continues, so CDI iteration order can no longer decide which services get reset on a partial failure.
- `AdminResetTest` now covers email too: sends a message via `/emails:send`, confirms it's captured, resets, and asserts `/emailMessages` is empty — closing the loop on both previously-skipped services.

251/251 locally.

PR #108, opened 2026-07-09
fix(acr): restart the shared registry container if it dies after first start

hectorvent:
## Summary

`AcrRegistryManager.ensureStarted()` was lazy-once: a `started` flag guarded the whole start path, so if the shared `registry:2` container died or was removed externally after a successful start, the manager kept handing out a dead endpoint. Every later registry create then sat in `provisioningState: Creating` forever (`isReady()` can never succeed), and azurerm/OpenTofu polled until their timeout.

Found live: an external cleanup removed `floci-az-acr-registry` mid-session and the OpenTofu compat suite timed out on `azurerm_container_registry`.

- `ensureStarted()` now verifies the container is actually running (`isContainerRunning`; `NotFoundException` → restart, transient inspect errors → assume alive so there are no restart storms) and starts a fresh container when it is gone.
- The readiness poller also re-ensures the registry, so a registry already stuck in `Creating` recovers instead of hanging forever.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

No wire-shape changes — same ARM responses, the `provisioningState` lifecycle just recovers to `Succeeded` the way it already does on first start. Matches the sidecar resilience rule (failures degrade gracefully, never wedge permanently). Verified against `hashicorp/azurerm` via the Terraform/OpenTofu compat suites.

## Checklist

- [x] `./mvnw test` passes locally (251/251)
- [ ] New or updated integration test added — the failure needs a real Docker container removed out-of-band mid-flight; validated instead by the OpenTofu compat suite against a native image with the exact repro

### DIFF
--- a/src/main/java/io/floci/az/services/acr/AcrHandler.java
+++ b/src/main/java/io/floci/az/services/acr/AcrHandler.java
@@ -356,10 +356,26 @@ public class AcrHandler implements AzureServiceHandler, Resettable {
         poller.scheduleAtFixedRate(() -> {
             try {
                 scanAll().forEach(reg -> {
-                    if ("Creating".equals(reg.getProvisioningState()) && registryManager.isReady()) {
-                        LOG.infov("ACR registry {0} is now ready", reg.getName());
-                        reg.setProvisioningState("Succeeded");
-                        putRegistry(reg.storageKey(), reg);
+                    if ("Creating".equals(reg.getProvisioningState())) {
+                        // Recovers the shared registry if it died after its initial start,
+                        // so a pending registry cannot stay Creating forever.
+                        registryManager.ensureStarted();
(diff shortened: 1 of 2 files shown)


hectorvent:
Good catch on the stale `loginServer` — applied in the amended commit, and went one step further than the suggestion:

- the `Creating` → `Succeeded` flip now refreshes `loginServer` from the live manager endpoint (as suggested), and
- already-`Succeeded` registries are converged too: if a recovery restart moved the shared registry to a new host port, the poller detects the drift and rewrites their stored `loginServer`, so records that were healthy *before* the container died don't keep advertising the dead port either.

251/251 locally.

PR #111, opened 2026-07-09
chore(core): extract shared ARM helpers into core/arm

hectorvent:
## Summary

The ARM-plane and sidecar services each carried private copies of the same path/JSON/error helpers, with real drift: `extractRg` existed in 10 variants across two fallback conventions, `parseBody` in 9, and 11 services hand-rolled the same CloudError JSON builder. This extracts a shared `core/arm` package — `ArmPaths` (path parsing with caller-supplied fallbacks), `ArmJson` (three body-parse variants + `cast`/`string`), `ArmErrors` (the `{"error":{"code","message"}}` builder), `ArmResources` (`_sub`/`_rg` keys + `stripInternal`) — and turns **57 service-local helper bodies across 17 services into one-line delegations**, so every call site keeps its exact historical semantics.

Behavior-preserving variant map:

- fallback `"unknown"`: arm, network, apim (ARM-plane convention)
- fallback `"default"`: redis, acr, aks, postgres, sql, vm (sidecar convention)
- `parseBody` lenient/immutable (`Map.of()` on miss): arm, network, apim, monitor, eventgrid
- `parseBody` lenient/mutable (`LinkedHashMap` on miss — callers mutate): keyvault, cosmos, appconfig
- `parseBodyStrict` (`400 InvalidRequestContent` semantics): no callers yet — landing pad for the planned ARM-plane strict-parse switch and the managed-identity feature branch
- untouched by design (different wire shapes): servicebus/eventhub flat `{"error":"msg"}` errors, cosmos data-plane errors, functions `AzureErrorResponse`, sidecar flat `methodNotAllowed`, apim's specialized `stripInternal`

Subtleties preserved verbatim and pinned by unit tests: `subscriptions` segment matching stays case-sensitive in `ArmHandler` and case-insensitive in sidecars; `resourceGroups` is case-insensitive everywhere. The fallback values are caller-supplied on purpose — services key their in-memory stores by them, so unifying `"unknown"`/`"default"` would silently change storage keys for malformed paths (a protocol decision, not a refactor).

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Azure Compatibility

Behavior-preserving by construction: the rewiring was applied by a script asserting exactly one match per method signature, each call site keeps its historical fallback/leniency, and the consolidated error builder emits the identical CloudError JSON (explicit `application/json` matches what JAX-RS already negotiated for Map entities). Verified with `./mvnw test` (263 incl. 12 new `ArmPaths`/`ArmJson` tests) plus the ARM-heavy compat suites — terraform, opentofu, azcli — all green on this exact head.

## Checklist

- [x] `./mvnw test` passes locally (263/263)
- [x] New or updated integration test added (`ArmPathsTest`, `ArmJsonTest` — pin the case-sensitivity and fallback semantics)

### DIFF
--- a/src/main/java/io/floci/az/core/arm/ArmErrors.java
+++ b/src/main/java/io/floci/az/core/arm/ArmErrors.java
@@ -1,0 +1,28 @@
+package io.floci.az.core.arm;
+
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+
+import java.util.Map;
+
+/**
+ * ARM CloudError responses — {@code {"error":{"code":"...","message":"..."}}}.
+ * The management-plane counterpart of {@code core/AzureErrorResponse} (which covers
+ * the storage data-plane XML/JSON shape).
+ */
+public final class ArmErrors {
+
+    private ArmErrors() {
+    }
+
+    public static Response error(int status, String code, String message) {
+        return Response.status(status)
+                .entity(Map.of("error", Map.of("code", code, "message", message)))
+                .type(MediaType.APPLICATION_JSON)
+                .build();
+    }
+
+    public static Response notFound(String message) {
+        return error(404, "ResourceNotFound", message);
+    }
+}
--- a/src/main/java/io/floci/az/core/arm/ArmJson.java
+++ b/src/main/java/io/floci/az/core/arm/ArmJson.java
@@ -1,0 +1,92 @@
+package io.floci.az.core.arm;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.floci.az.core.AzureRequest;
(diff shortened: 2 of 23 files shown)


hectorvent:
All three applied in the amended commit:

- `parseBodyMutable` now guards a null body stream (returns a fresh mutable map, matching its siblings) — pinned by a new assertion in `ArmJsonTest`. Good catch: the gap existed in the original per-service copies too and was preserved by the behavior-preserving extraction, but a shared utility is the right place to close it.
- `ArmPaths.resourceName` caches compiled patterns in a `ConcurrentHashMap` keyed by resource type (a small fixed set), so hot-loop callers don't pay a per-call compile.
- All three parse variants capture `req.bodyStream()` once into a local instead of re-reading the accessor.

PR #112, opened 2026-07-09
feat(network): add Private Endpoint and Private DNS support

### LINKED ISSUE #60: Add Azure Private Endpoint and Private DNS support
## Title

Add Azure Private Endpoint and Private DNS support

## Description

Implement Azure Private Endpoint and Private DNS Zone support inside floci-az.

The purpose is to complete the Security & Governance parity set and let Terraform / SDK clients that declare Private Link + private DNS plumbing apply cleanly against the emulator. There is no real network isolation to enforce locally — this is mostly ARM metadata plus synthesized private IPs and DNS records.

ARM providers: `Microsoft.Network/privateEndpoints`, `Microsoft.Network/privateDnsZones` (+ `virtualNetworkLinks`, `A`/`CNAME` record sets), and `privateLinkServices`.

Suggested capabilities:

* Private endpoint CRUD with `privateLinkServiceConnections` + synthesized NIC/private IP
* Private DNS zone CRUD + virtual network links
* Private DNS record sets (A / CNAME / SOA)
* `privateDnsZoneGroups` on the endpoint

floci-az notes: extend the existing `network` / `ArmHandler` `Microsoft.Network` stubs; low complexity, mostly metadata. Pairs naturally with the Azure DNS issue.

## Official Documentation

* https://learn.microsoft.com/en-us/azure/private-link/
* https://learn.microsoft.com/en-us/azure/dns/private-dns-overview
* https://learn.microsoft.com/en-us/rest/api/virtualnetwork/private-endpoints

## Azure SDK for Java

* Management: `com.azure.resourcemanager:azure-resourcemanager-network` — https://learn.microsoft.com/en-us/java/api/overview/azure/resourcemanager-network-readme

## Azure REST API Specs

* https://github.com/Azure/azure-rest-api-specs/tree/main/specification/network
* https://github.com/Azure/azure-rest-api-specs/tree/main/specification/privatedns

hectorvent:
## Summary

Adds Private Endpoint and Private DNS emulation to the Microsoft.Network provider:
`privateEndpoints`, `privateDnsZones` (with `virtualNetworkLinks` and
A/AAAA/CNAME/MX/PTR/SOA/SRV/TXT record sets), `privateLinkServices`, and
`privateDnsZoneGroups`. Private DNS zones seed a default SOA record and enforce
ETag concurrency with record/link counters; private endpoints get auto-approved
connections, a synthesized NIC, zone groups, and cascade delete; private link
services expose a synthesized alias.

Closes #60

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Verified against the Azure SDK for Java (BOM 1.2.28, `azure-resourcemanager-network`
/ `azure-resourcemanager-privatedns`) via `PrivateLinkCompatibilityTest`, and against
the `hashicorp/azurerm` Terraform provider (`~> 3.0`) plus OpenTofu with new
`azurerm_private_dns_zone`, `azurerm_private_dns_zone_virtual_network_link`,
`azurerm_private_endpoint`, and record-set resources in both BATS suites.

## Checklist

- [x] `./mvnw test` passes locally (267/267)
- [x] New or updated integration test added (`NetworkHandlerTest` +249 lines,
      `PrivateLinkCompatibilityTest` 267 lines, Terraform/OpenTofu BATS assertions)


hectorvent:
Thanks — all three applied in the amended commit:

**NIC leak (the P1)** — you're right, this was a real corruption path. `createOrUpdatePrivateEndpoint` now reuses the endpoint's existing NIC on update instead of minting a fresh randomly-named one each PUT (Azure keeps a single stable NIC for the endpoint's lifetime). A NIC is only synthesized on first create, or when an explicit `customNetworkInterfaceName` actually changes it — in which case the replaced NIC is removed too, so there's no orphan either way. New regression test `privateEndpointUpdateReusesNicAndDoesNotLeak`: create + two updates → exactly one NIC in the listing, stable id, and cascade delete leaves zero.

**201 on create (P2)** — done, matching `createOrUpdatePrivateDnsZone` and the ARM contract. (Updated the existing lifecycle test; the Java compat suite uses a 2xx `assertOk` so it's unaffected.)

**Zone ETag rotation (P2)** — done, now regenerated on every write like the record sets and VNet links already do.

**Trailing-slash guard (nit)** — addressed, but not by just dropping the slash: `handlePrivateEndpoints` had no bare-list branch, so `GET .../privateEndpoints` would've 404'd instead of falling through. Added an explicit `parts.length == 1` list branch and then dropped the slash, so it's consistent with `privateDnsZones` and no longer depends on the generic fallthrough.

268/268 unit tests green.

PR #113, opened 2026-07-09
feat(managedidentity): add Managed Identity emulation (ARM CRUD + IMDS token endpoint)

hectorvent:
## Summary

Adds `Microsoft.ManagedIdentity` + the IMDS token endpoint, HTTP-only (no sidecar):

- **ARM CRUD** for user-assigned identities (server-generated
  `principalId`/`clientId`/`tenantId` GUIDs, stable across updates; PUT is
  full-replace for tags per msi spec 2024-11-30), `federatedIdentityCredentials`
  children, and the system-assigned `GET {scope}/…/identities/default` read with
  deterministic per-scope GUIDs. Identities appear in the resource group's
  `/resources` listing (azurerm's pre-delete emptiness check). Malformed bodies
  answer `400 InvalidRequestContent`.
- **IMDS token endpoint** (`GET /metadata/identity/oauth2/token`, imds spec
  2023-07-01, GET-only, `Metadata: true` required): resolves an identity by
  `client_id`/`object_id`/`msi_res_id` or synthesizes the system-assigned one,
  and returns the all-string response with a v1.0 JWT (`appid`/`oid`/`idtyp=app`)
  signed by the Entra key — verifiable against the emulator JWKS. The
  system-assigned scope is configurable (`services.managed-identity.system-assigned-scope`)
  so token `oid` claims can match `identities/default` reads.
- Built on the shared plumbing this repo just landed: `Resettable`, `core/arm`
  (`parseBodyStrict`), `RequestUrls`, `TokenIssuer.deterministicGuid`; suite env
  flows through `POD_IDENTITY_ENV` → `SUITE_ENV_*` and the CI matrix.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Shapes verified against the local msi 2024-12-30 / imds 2023-07-01 specs.
Verified end-to-end with `azure-identity` `ManagedIdentityCredential` in **Java,
Python, and Node.js** (via `AZURE_POD_IDENTITY_AUTHORITY_HOST`) — all three
suites include an SDK-level acquisition test. Intentional deviations documented
in `docs/services/managed-identity.md`.

## Checklist

- [x] `./mvnw test` passes locally (281/281)
- [x] New or updated integration test added (`ManagedIdentityTest` 15 tests +
      `ManagedIdentityDisabledTest`, plus Java/Python/Node compat suites)

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -73,12 +73,17 @@ jobs:
       matrix:
         include:
           - test: sdk-test-python
+            extra_env: >-
+              -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
           - test: sdk-test-java
             extra_env: >-
               -e SERVICEBUS_HOST=floci-az-servicebus-default
               -e SERVICEBUS_AMQPS_PORT=5671
               -e SERVICEBUS_NAMESPACE=default
+              -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
           - test: sdk-test-node
+            extra_env: >-
+              -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
           - test: compat-terraform
           - test: compat-opentofu
           - test: compat-azcli
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -233,12 +233,13 @@ make stop
 ```
 compatibility-tests/
   sdk-test-python/      — Azure SDK for Python (pytest): blob/queue/table, cosmos, vm, redis, acr,
-                          plus appconfig/ and keyvault/ sub-suites
+                          managed identity, plus appconfig/ and keyvault/ sub-suites
   sdk-test-java/        — Azure SDK for Java (BOM; the primary compat target): blob/queue/table,
                           cosmos + all engines, functions, keyvault, appconfig, eventhub,
-                          servicebus, eventgrid, apim, sql, postgres, vm, datalake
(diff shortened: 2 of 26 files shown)


hectorvent:
Thanks for the careful pass. Fixed the two correctness issues; on the two security notes I'll explain rather than change, with reasoning below.

**PATCH live-map mutation (the P1)** — fixed. `patchIdentity` now copies the existing map, mutates the copy, and does an atomic `store.putIdentity(key, updated)` replace instead of calling `existing.put(...)` on the map held inside the ConcurrentHashMap. Good catch — CHM guards its structure, not the LinkedHashMap values.

**`removeIdentity` FIC-cleanup ordering (P2)** — applied your suggestion: clean the FICs first, then remove the identity, so a re-create in the gap can't have its fresh FICs swept.

**Host header → `iss` / `clientSecretUrl` (the two security P2s)** — these are intentional and I'm leaving them as-is. `RequestUrls.resolveBaseUrl` deriving the base URL from the `Host` header is the emulator-wide pattern (shared with `EntraServiceHandler`): it's what keeps the issuer, JWKS, discovery document, and generated URLs consistent no matter how the emulator is reached — localhost, the `floci-az:4577` container name the compat suites actually use, or a `.vault.azure.net`-style host. An allowlist that rejected non-localhost Hosts would break those SDK/compat flows. It's dev-only localhost tooling, and when a fixed issuer is wanted, `floci-az.services.entra.issuer` already overrides the Host-derived value. Happy to add a doc note calling this out explicitly if useful.

281/281 unit tests green.

PR #130, opened 2026-07-16
feat(servicebus): subscription rules and topic filters (CorrelationFilter, SqlFilter, $Default)

### LINKED ISSUE #124: [FEAT] Service Bus: subscription rules/filters (CorrelationFilter, SqlFilter, $Default rule management)
## Service

Service Bus

## API Action / Feature

Subscription rules: the `/{topic}/subscriptions/{sub}/rules` management-plane endpoints (create/list/delete), `CorrelationFilter` and `SqlFilter` evaluation on the data plane, and Azure's implicit `$Default` TrueFilter rule semantics (auto-created accept-all rule; first user rule typically replaces it via delete-then-add).

As of `docs/services/service-bus.md` the management plane covers entity CRUD for queues/topics/subscriptions, but there are no rule endpoints, and a code search for `CorrelationFilter` returns no hits — so today every subscription receives the whole topic.

## Azure Documentation

https://learn.microsoft.com/en-us/azure/service-bus-messaging/topic-filters

## Why is this needed?

We evaluated floci-az for a .NET Aspire app with a shared-topic event architecture: multiple consumer services subscribe to the same topics and each subscription carries a `CorrelationFilter` (on the message subject/label) so a consumer only receives the event types it handles. We consolidated ~75 subscriptions down to ~37 by relying on these filters.

Without rule support, every consumer receives every event on the topic — that changes delivery semantics (wrong-consumer deliveries, inflated delivery counts, spurious dead-lettering once MaxDeliveryCount is exceeded on messages a consumer was never meant to see).

Filter evaluation could potentially map to Artemis filter expressions (SQL92 selectors) for the `SqlFilter` case, with `CorrelationFilter` compiled down to a selector over the mapped AMQP properties.

- [ ] Yes
- [x] No (happy to help test against a real-world multi-subscription topology)

thomhurst:
Closes #124

## What

Subscription rules / topic filters for Service Bus: the `/{topic}/subscriptions/{sub}/rules[/{rule}]` management-plane endpoints, Azure's implicit `$Default` TrueFilter rule, and broker-side filter evaluation (CorrelationFilter, SqlFilter, TrueFilter, FalseFilter) by compiling rules to Artemis SQL92 queue selectors.

## Management plane

- `PUT` / `GET` / `DELETE` `/{topic}/subscriptions/{sub}/rules/{rule}` and `GET .../rules` (ATOM `RuleDescription` wire format), on both the spec paths used by `ServiceBusAdministrationClient` and the legacy `/{ns}/topics/...` paths.
- Every new subscription gets the implicit `$Default` TrueFilter rule. Both standard SDK flows work: add-rule-then-delete-`$Default`, and `DefaultRuleDescription` embedded in the subscription create body (`CreateSubscriptionAsync(subscriptionOptions, ruleOptions)` — the flow the issue's Aspire topology uses).
- Rule PUT upserts (201 create / 200 replace); cascade deletes on subscription and topic removal.

## Data plane

Filters compile to Artemis core filters on the subscription's MULTICAST queue, so evaluation happens **inside the broker at routing time** — non-matching messages are never routed to the subscription: no wrong-consumer deliveries, no delivery-count inflation, no spurious dead-lettering (the exact failure modes described in #124).

Identifier mapping (verified against Artemis `AMQPMessage.getObjectProperty`):

| Azure | Artemis selector |
|---|---|
| `CorrelationId` / `sys.CorrelationId` | `JMSCorrelationID` (AMQP correlation-id) |
| `Label` / `Subject` / `sys.Label` / `sys.Subject` | `JMSType` (AMQP subject) |
| `SessionId` / `sys.SessionId` | `JMSXGroupID` (AMQP group-id) |
| application properties | referenced by name |

- SqlFilter translation strips `user.`/`[bracket]` forms and rewrites `EXISTS(p)` → `(p IS NOT NULL)`; `LIKE`, `IN`, `BETWEEN`, `IS NULL`, arithmetic and boolean operators pass through.
- Correlation property values typed `int`/`long`/`double`/`boolean` compare with their declared type (JMS selectors never match across types).
- Multiple rules OR-combine into a single delivery; a subscription with no rules receives nothing (`1=0`), matching Azure.
- Rule changes update the queue filter **in place** via `ActiveMQServerControl.updateQueue` — routed messages and attached receivers survive; changes apply to future messages only, as on Azure.

## Documented deviations

- Filters on `MessageId`, `To`, `ReplyTo`, `ReplyToSessionId`, `ContentType` are rejected with HTTP 400 — these AMQP fields have no broker-side selector identifier, so rejecting loudly beats silently misrouting. (Fixable later via a small Artemis broker-plugin that copies them into filterable properties.)
- `SqlRuleAction` is stored and echoed by the management API but not applied to delivered messages (no per-subscription transformer without a divert-based re-architecture).

Both are documented in `docs/services/service-bus.md`.

## Also fixed

- Service Bus ATOM feed responses embedded a full `<?xml?>` prolog inside every `<entry>`, making list responses malformed XML for strict parsers. Entry builders are now prolog-free; responses prepend it once.

## Testing

- **67 new unit tests** across selector compilation (26), RuleDescription XML codec (19), and REST rule CRUD (22) — happy paths plus non-happy: 400s leave no partial state, 404s, 405 on non-GET rules collection, malformed-body fallbacks, cascade behavior, prolog regression. Full suite: 376 green.
- **End-to-end against a real Artemis sidecar** with real SDK traffic — 7 scenarios verified (Label routing, typed int property, SQL filter with `sys.Label` + numeric comparison, OR rules single-copy, no-rules-delivers-nothing, SessionId routing, `$Default` accepts-all), and compiled selectors inspected on the broker via Jolokia.
- `ServiceBusRuleCompatibilityTest` added to the Java SDK compat suite (same skip-when-mocked behavior as the existing Service Bus AMQP tests).

## Pre-existing issue found (not addressed here)

In real-broker mode, **every** Service Bus Java SDK send currently fails with `"Size of the payload exceeded maximum message size: 0 kb"`: azure-core-amqp leaves its link size at 0 when the broker's ATTACH carries no `max-message-size`, and Artemis never advertises one (verified on 2.42.0 and 2.44.0 — no config knob exists). CI never sees this because `mocked: true` skips all AMQP compat tests. Worth its own issue; e2e verification for this PR used a different SDK stack to sidestep it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **servicebus:** subscription rules and topic filters — the `/{topic}/subscriptions/{sub}/rules[/{rule}]` management-plane endpoints (create/get/list/delete, ATOM `RuleDescription` wire format with `CorrelationFilter`/`SqlFilter`/`TrueFilter`/`FalseFilter` and `SqlRuleAction` bodies), Azure's implicit `$Default` TrueFilter rule (auto-created per subscription, replaceable via the delete-then-add SDK flow or a `DefaultRuleDescription` in the subscription create body), and broker-side filter evaluation: rules compile to Artemis SQL92 queue selectors (`CorrelationId`→`JMSCorrelationID`, `Label`/`Subject`→`JMSType`, `SessionId`→`JMSXGroupID`, application properties by name, with typed int/long/double/boolean correlation values), multiple rules OR-combine into a single delivery, no rules delivers nothing, and rule changes update the queue filter in place (`updateQueue`) without dropping routed messages or kicking receivers. Filters on `MessageId`/`To`/`ReplyTo`/`ReplyToSessionId`/`ContentType` are rejected with 400 (no broker-side AMQP mapping); `SqlRuleAction` is stored/echoed but not applied to delivered messages ([#124](https://github.com/floci-io/floci-az/issues/124))
+
+### Fixed
+
+- **servicebus:** ATOM feed responses (queues/topics/subscriptions list) no longer embed an XML prolog inside every `<entry>`, which made the feed malformed XML for strict parsers
+
 ## [0.9.0] - 2026-07-09
 
 ### Added
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/EmulatorConfig.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/EmulatorConfig.java
@@ -502,6 +502,50 @@ public final class EmulatorConfig {
     }
 
     /**
+     * Creates or replaces a subscription rule via the floci-az management API.
+     * {@code ruleDescriptionXml} is the inner {@code <RuleDescription>} element
+     * (with xmlns declarations), as sent by ServiceBusAdministrationClient.
+     */
+    static void ensureServiceBusRule(String topicName, String subName,
+                                      String ruleName, String ruleDescriptionXml) throws Exception {
+        String url = BASE + "/" + ACCOUNT + "-servicebus/" + SERVICEBUS_NAMESPACE
+                + "/topics/" + topicName + "/subscriptions/" + subName + "/rules/" + ruleName;
+        String body = "<entry xmlns=\"http://www.w3.org/2005/Atom\"><content type=\"application/xml\">"
+                + ruleDescriptionXml + "</content></entry>";
+        byte[] bodyBytes = body.getBytes(StandardCharsets.UTF_8);
+        HttpURLConnection conn = (HttpURLConnection) URI.create(url).toURL().openConnection();
+        conn.setRequestMethod("PUT");
+        conn.setDoOutput(true);
+        conn.setConnectTimeout(5_000);
+        conn.setReadTimeout(30_000);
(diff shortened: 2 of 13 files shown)


hectorvent:
Thank you, this is an impressive piece of work, and evaluating filters inside the broker at routing time is the right architecture for exactly the failure modes from #124. I compared the ATOM wire shape element by element against the Azure SDK for Java's own serializers ([RuleDescription](https://github.com/Azure/azure-sdk-for-java/blob/ddf934e/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/administration/implementation/models/RuleDescription.java), [CorrelationFilter](https://github.com/Azure/azure-sdk-for-java/blob/ddf934e/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/administration/implementation/models/CorrelationFilter.java)) and it matches throughout, including both SDK rule-creation flows and your documented deviations.

One small suggestion: `ServiceBusRuleXml.parse` catches `Exception` and returns empty without logging, and the TrueFilter fallback then turns a malformed SqlFilter body into an accept-everything rule with no trace. A `LOG.debugf` with the parse failure would make that diagnosable (AGENTS.md asks that tolerated exceptions still be logged). Not blocking.

Heads-up, not for this PR: CI currently skips both Service Bus compat suites silently (`Tests run: 0`, a repo infrastructure gap reproducible on main), so the green check is not broker-side coverage yet.

Looks good to merge from my side.

PR #133, opened 2026-07-17
fix(storage): serialize service properties XML instead of returning toString

### LINKED ISSUE #131: [BUG] Queue: Get Queue Service Properties returns Java toString instead of XML
## Description

`GET /{account}-queue?restype=service&comp=properties` (Get Queue Service Properties) returns the Java object identity string of the XML builder instead of the serialized XML document. Azure SDK clients (e.g. `Azure.Storage.Queues` `QueueServiceClient.GetPropertiesAsync`) fail with an XML parse error.

## Reproduction

```bash
docker run --rm -p 4577:4577 floci/floci-az:0.9.0
curl -i "http://localhost:4577/devstoreaccount1-queue?restype=service&comp=properties"
```

## Actual

```
HTTP/1.1 200 OK
content-length: 36
Content-Type: application/xml

io.floci.az.core.XmlBuilder@49baf871
```

Looks like the response writer calls `toString()` on the builder instead of serializing it.

## Expected

A `<StorageServiceProperties>` XML document, per https://learn.microsoft.com/en-us/rest/api/storageservices/get-queue-service-properties

## Impact

.NET SDK throws:

```
System.Xml.XmlException : Data at the root level is invalid. Line 1, position 1.
   at Azure.Storage.Queues.ServiceRestClient.GetPropertiesAsync(...)
```

## Environment

- floci/floci-az:0.9.0 (sha256:5e403a7d788c...) on Docker Desktop, Windows 11
- Found while adding a floci-az module to testcontainers-dotnet

thomhurst:
## Summary

Fixes #131.

`GET /{account}-queue?restype=service&comp=properties` returned the Java object identity string (`io.floci.az.core.XmlBuilder@...`) instead of the serialized `<StorageServiceProperties>` document, because the handler called `toString()` on `XmlBuilder`, which only exposes `build()` and never overrode `Object.toString()`. Azure SDK clients failed with `System.Xml.XmlException: Data at the root level is invalid`.

The Blob service properties endpoint (`getBlobServiceProperties`) had the identical bug and is fixed here too.

## Changes

- `QueueServiceHandler.getQueueServiceProperties` and `BlobServiceHandler.getBlobServiceProperties`: call `.build()` instead of `.toString()`.
- `XmlBuilder`: override `toString()` to delegate to `build()`, so any future accidental string conversion produces the XML rather than the object identity string.
- Regression tests in `QueueServiceTest` and `BlobServiceTest` asserting the endpoints return 200 with a document starting `<StorageServiceProperties>`.

## Testing

Full test suite passes locally (`mvnw test`), including the two new regression tests.

### DIFF
--- a/src/main/java/io/floci/az/core/XmlBuilder.java
+++ b/src/main/java/io/floci/az/core/XmlBuilder.java
@@ -108,6 +108,12 @@ public final class XmlBuilder {
         return sb.toString();
     }
 
+    /** Same as {@link #build()}, so accidental string conversion never leaks the object identity. */
+    @Override
+    public String toString() {
+        return build();
+    }
+
     /**
      * Escapes the five XML special characters: {@code & < > " '}.
      * Returns an empty string for null or empty input.
--- a/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
+++ b/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
@@ -187,7 +187,7 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
                 .end("MinuteMetrics")
                 .start("StaticWebsite").elem("Enabled", "false").end("StaticWebsite")
             .end("StorageServiceProperties")
-            .toString();
+            .build();
         return Response.ok(xml, "application/xml").build();
     }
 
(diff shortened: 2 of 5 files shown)


hectorvent (approved):
Thanks @thomhurst 
Solid fix

PR #135, opened 2026-07-17
fix(postgres): answer 202 with Location so azurerm 3.x and 4.x both apply

Ali-Shaikh:
## Summary

Revised per review: **202 is the only status both provider majors accept**, and the fix has to cover the child resources too, not just the server.

Every mutating operation on Flexible Servers (and their databases, firewall rules and configurations) now returns `202 Accepted` with a `Location` header pointing at the resource. `DELETE` is unchanged at `204`.

## Why 202, and why the child resources

I checked the vendored `go-azure-sdk` in both provider majors:

| Operation | azurerm 3.x accepts | azurerm 4.x accepts |
|---|---|---|
| `PUT` (create/update) | 200, 201, 202 | **202 only** |
| `PATCH` (update) | 200, 202 | **202 only** |
| `DELETE` | 200, 202, 204 | 202, 204 |

Two things follow. **202 is the only value in the intersection**, which is why the previous revision of this PR (200) traded one incompatibility for another, exactly as you said. And the same narrowing applies to `databases`, `firewallRules` and `configurations`, so the original scope was insufficient: fixing only the server would have moved the failure to the very next resource in the config.

This also explains why CI never caught it. `compat-opentofu/provider.tf` pins `~> 3.0`, and 3.x accepts the 201 that main returns.

Because the emulator has already finished the work when it answers, the `Location` header points back at the resource itself. The SDK's poller then reads the terminal `provisioningState` from a GET of that URL and completes on its first poll. The metadata-only children carry no `provisioningState`, which the poller explicitly treats as completion, so they need no body changes.

## On the version question

You asked which provider version reproduced the original 201 error, and I have to admit that session didn't record it, so I can't answer honestly for the original report. Everything in this revision is pinned and reproducible: **azurerm 4.81.0** for the failing case, and the suite's own `~> 3.0` constraint for the regression check.

## Validation

Run against a real emulator container (`floci-az:test` on the compat network, TLS on, real Postgres sidecar), driving the actual provider rather than asserting status codes in isolation:

- **Before, azurerm 4.81.0, unpatched `main`:** fails exactly as reported.
  ```
  Error: creating Flexible Server (...): performing CreateOrUpdate:
  unexpected status 201 (201 Created) with response: {... "provisioningState":"Succeeded" ...}
  ```
- **After, azurerm 4.81.0:** `Apply complete! Resources: 4 added, 0 changed, 0 destroyed.` A follow-up `plan` reports no differences (no perpetual diff), and `destroy` removes all four.
- **Regression check, the repo's OpenTofu suite at `~> 3.0`:** 20/20 green, including the postgres server and database cases.
- **Unit and HTTP suites:** 570 tests, 0 failures, with the Docker-backed `PostgresDockerTest` actually executing against a real container. `PostgresHandlerMockedTest` gains two tests that replay the poller's exchange: PUT returns 202 with a `Location`, and a GET of that URL reports `Succeeded`.

## Note on scope

`SqlHandler` returns 201 on create and looks to have the identical problem for 4.x users. I've left it out of this PR to keep the change reviewable, and I'm happy to raise it separately or fold it in, whichever you prefer.

The 202 shape also makes a genuinely asynchronous create possible later (answer immediately with `provisioningState: Creating` and let the poller wait), which would address the cold-start complaint in #138. That felt like a separate change, so this PR keeps create synchronous.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-
-- **postgres:** mutating operations on Flexible Servers (and their databases, firewall rules and configurations) now answer `202 Accepted` with a `Location` header pointing at the resource, instead of `200`/`201`. ARM models these as long-running operations and the provider's accepted status codes narrowed between majors: azurerm 3.x accepts 200/201/202 on PUT, but **4.x accepts only 202**, so every `azurerm_postgresql_flexible_server*` apply failed against the emulator with `unexpected status 201 (201 Created)`. 202 is the only value both majors accept. Because the work is already complete when we answer, the `Location` header points back at the resource itself and the SDK's poller finishes on its first poll. `DELETE` is unchanged at 204, which both majors already accept ([#135](https://github.com/floci-io/floci-az/pull/135))
-
 ## [0.10.0] - 2026-07-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **postgres:** mutating operations on Flexible Servers (and their databases, firewall rules and configurations) now answer `202 Accepted` with a `Location` header pointing at the resource, instead of `200`/`201`. ARM models these as long-running operations and the provider's accepted status codes narrowed between majors: azurerm 3.x accepts 200/201/202 on PUT, but **4.x accepts only 202**, so every `azurerm_postgresql_flexible_server*` apply failed against the emulator with `unexpected status 201 (201 Created)`. 202 is the only value both majors accept. Because the work is already complete when we answer, the `Location` header points back at the resource itself and the SDK's poller finishes on its first poll. `DELETE` is unchanged at 204, which both majors already accept ([#135](https://github.com/floci-io/floci-az/pull/135))
+
 ## [0.10.0] - 2026-07-31
 
 ### Added
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/PostgresCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/PostgresCompatibilityTest.java
@@ -84,8 +84,10 @@ class PostgresCompatibilityTest {
         HttpResponse<String> resp = send("PUT",
             ARM_BASE + "/flexibleServers/" + SERVER + API, serverBody, Duration.ofMinutes(5));
 
-        assertTrue(resp.statusCode() == 200 || resp.statusCode() == 201,
-            "PUT server failed (" + resp.statusCode() + "): " + resp.body());
+        // 202 + Location: the ARM long-running-operation shape both azurerm 3.x and 4.x accept.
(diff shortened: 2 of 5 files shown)


hectorvent:
Thank you for chasing this down, the write-up of the failure mode is spot on. One catch before merging: I think 200 unfortunately trades one incompatibility for another.

The current provider's `CreateOrUpdate` client accepts only **202** ([method_createorupdate.go](https://github.com/hashicorp/terraform-provider-azurerm/blob/e6c4fe66/vendor/github.com/hashicorp/go-azure-sdk/resource-manager/postgresql/2025-08-01/servers/method_createorupdate.go), `ExpectedStatusCodes: []int{http.StatusAccepted}`), and the 2025-08-01 REST spec likewise defines 202 as the sole success response for the Flexible Server PUT ([openapi.json](https://github.com/Azure/azure-rest-api-specs/blob/ced4668607/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2025-08-01/openapi.json)). So a 200 fails the same way the 201 does, just with a different number in the error. Which provider version did you reproduce against? I suspect an earlier 4.x minor.

Suggested shape: return **202 Accepted** with the JSON body plus a `Location` header pointing back at the server URL itself. go-azure-sdk then polls provisioningState via GET and finishes on the first poll, and the older 3.x provider handles the same flow, so it satisfies both. Happy to help test a revised version, and thanks again for the groundwork here.

hectorvent:
Thank you for turning this around so thoroughly. The 202 with a self referencing `Location` is exactly right, and the javadoc table of what azurerm 3.x and 4.x each accept is genuinely useful for whoever reads this next.

I re verified the shape against both sources: the provider's client accepts only 202 ([method_createorupdate.go](https://github.com/hashicorp/terraform-provider-azurerm/blob/da2e1b29/vendor/github.com/hashicorp/go-azure-sdk/resource-manager/postgresql/2025-08-01/servers/method_createorupdate.go), `ExpectedStatusCodes: []int{http.StatusAccepted}`), and the 2025-08-01 spec declares 202 as the sole success response for the Flexible Server PUT. Leaving DELETE at 204 is right too, the spec allows 202 and 204 there.

**(blocking)** One thing left, and it is not your doing: `CHANGELOG.md` became automation generated on main after your last push, so that hunk now conflicts with the release commit and the new guard fails on it. Would you drop it? Your change is still credited automatically, the release notes are generated from the PR title.

That is the only blocker from my side.

Ali-Shaikh:
Dropped the CHANGELOG.md hunk in db11db5. File matches main again so changelog-guard doesn't fire on this PR, no label needed. Rest of the fix is untouched, all checks green.

hectorvent:
Thank you, and sorry for the extra round trip on the changelog. This is ready from my side.

I went back through the implementation properly this time rather than just the status codes. Preserving the caller's `api-version` on the `Location` URL is a good catch: the poller then reads back the same version it wrote to. And the opentofu suite actually provisions a flexible server, a database and a firewall rule against azurerm 3.x, so the "3.x accepts 202 too" half of your javadoc table is confirmed by a live run, not just by reading the 4.x client.

**(follow-up, separate PR)** `resourceLocation()` builds the URL from the `Host` header, where `VmHandler` and five other services use `EmulatorConfig.effectiveBaseUrl()`. They agree in practice today, but a configured base URL override would be honoured by one and not the other. Worth aligning at some point, not worth holding this.

No blockers from my side.

hectorvent (approved):
Verified against the 2025-08-01 spec and the vendored azurerm client, and the opentofu suite confirms azurerm 3.x accepts the same shape. Approving.

Issue #138, opened 2026-07-18
[PERF] Azure SQL: avoid 2.29 GB first-provision cold start


thomhurst:
## Summary

Floci-AZ itself starts in milliseconds, but the first Azure SQL logical-server provisioning request blocks while Floci-AZ downloads and starts a full SQL engine image.

In a clean environment, this pulled `mcr.microsoft.com/azure-sql-edge:latest`, reported by Docker as a **2.29 GB** image, and kept the ARM `PUT` request open for approximately **3 minutes 12 seconds**. Clients with ordinary HTTP timeouts report failure even though the server eventually becomes `Ready`.

This issue proposes separating fast Azure SQL control-plane emulation from the optional heavyweight SQL data plane. It does **not** propose reimplementing SQL Server or T-SQL.

## Environment

- Floci-AZ: `0.9.0`
- Container image: `floci/floci-az`
- SQL engine image selected by Floci-AZ: `mcr.microsoft.com/azure-sql-edge:latest`
- Docker-reported SQL engine image size: `2.29 GB`
- SQL EULA enabled with `FLOCI_AZ_SERVICES_SQL_ACCEPT_EULA=Y`
- Docker socket mounted so Floci-AZ could provision the engine container

## Reproduction

Start Floci-AZ with Azure SQL enabled:

```bash
docker run --rm \
  --name floci-az \
  -p 4577:4577 \
  -e FLOCI_AZ_SERVICES_SQL_ACCEPT_EULA=Y \
  -v /var/run/docker.sock:/var/run/docker.sock \
  floci/floci-az
```

From a clean Docker image cache, create a logical SQL server through the ARM contract:

```bash
curl --request PUT \
  'http://localhost:4577/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/floci-local/providers/Microsoft.Sql/servers/repro-sql?api-version=2021-11-01' \
  --header 'content-type: application/json' \
  --data '{
    "location": "uksouth",
    "properties": {
      "administratorLogin": "sqladmin",
      "administratorLoginPassword": "StrongPassw0rd!"
    }
  }'
```

## Observed timeline

The following timestamps came from one first-provision run:

| Time | Event |
| --- | --- |
| `20:17:04` | ARM `PUT` received |
| `20:17:04` | Pull started for `mcr.microsoft.com/azure-sql-edge:latest` |
| `20:20:06` | SQL Server TCP port opened; engine initialization still running |
[...]

PR #139, opened 2026-07-18
fix(sql): replace retired Azure SQL Edge image

thomhurst:
## Summary

- Replace retired `mcr.microsoft.com/azure-sql-edge:latest` with `mcr.microsoft.com/mssql/server:2025-latest`.
- Remove Edge-only `SA_PASSWORD`; retain supported `MSSQL_SA_PASSWORD`.
- Fail startup and clean up the sidecar when its TDS port never opens.
- Document SQL Server container architecture support and update examples, tests, and changelog.

## Why

Azure SQL Edge retired on September 30, 2025. Floci-AZ should use Microsoft's supported SQL Server container for its real TDS data plane.

## Validation

- `docker pull mcr.microsoft.com/mssql/server:2025-latest` succeeded.
- Focused SQL unit suite: 12 tests passed.
- Real `SqlCompatibilityTest` against a Floci-managed SQL Server 2025 container: 11/11 passed.
- Checks rerun after this push.

Refs #138


hectorvent:
@thomhurst please rebase main to fix actions failures.

hectorvent:
Thank you, this is a clean and complete swap: config default, YAML, docs, README and the `SA_PASSWORD` removal all line up, and running the compat suite against a live SQL Server 2025 container is exactly the validation this needed.

I read the two commits you added since my last look. Failing the start when the container never becomes ready is right, and tearing down the unready container rather than leaving it behind is a good catch. I checked it against the rule that sidecar startup must not be fatal, and it is fine: this path is only reached from a provisioning request, which already turns the failure into a 500 rather than taking the emulator down.

**(blocking)** The one thing left is `CHANGELOG.md`. That file is generated by semantic release from commit messages, so hand edits conflict with the release commit and the guard fails on it. Would you drop that hunk? Your change is still credited automatically, the release notes come from the PR title.

thomhurst:
Dropped the `CHANGELOG.md` hunk and pushed the update. The focused SQL suite is green (9 tests).

hectorvent:
Thank you, that is the changelog hunk sorted and this is good to go from my side.

For the record on the Compatibility check that went red after your last push: that was the python queue visibility timeout test, which has nothing to do with your SQL image swap. It sends a message with a one second visibility timeout and then immediately asserts it is not visible, so it flips whenever the runner is slightly slow, and it failed on an unrelated branch the same day. It has since been rerun and passed, so everything is green here now.

No blockers from my side.

hectorvent (approved):
Changelog hunk dropped as asked. The earlier red check was the flaky python queue test and has since passed on rerun. Approving.

PR #141, opened 2026-07-19
feat(cosmos): container DefaultTimeToLive (TTL) with per-document overrides

### LINKED ISSUE #126: [FEAT] Cosmos DB: container DefaultTimeToLive (TTL) support
## Service

Cosmos DB (SQL API)

## API Action / Feature

Container-level `defaultTtl` (accepted on container create/replace) plus per-document `ttl` override, with expired documents excluded from reads/queries and eventually purged — matching Azure's TTL semantics (`defaultTtl: -1` = on but no default, `n` = n seconds, absent = off).

Today container create supports partition key path + default indexing policy only; a code search for `defaultTtl`/`timeToLive` in the cosmos service returns no hits.

## Azure Documentation

https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/time-to-live

## Why is this needed?

We evaluated floci-az for a .NET Aspire app that uses Cosmos for chat and email-audit storage. Several containers declare `DefaultTimeToLive` at provision time (e.g. sent-email logs with a retention window). With TTL silently ignored, local behavior diverges from production in a way tests can't see: retention/compliance logic that depends on expiry appears to "work" locally (documents never disappear) and only misbehaves in prod.

Even a simplified implementation (filter expired docs at read/query time based on `_ts` + ttl, lazy purge) would restore parity for dev/test purposes — the background-purge timing isn't the part apps depend on.

- [ ] Yes
- [x] No

Ali-Shaikh:
## Summary

Implements container `DefaultTimeToLive` for the Cosmos DB SQL API emulator (closes #126).

- `defaultTtl` is accepted and validated on container create (`-1` or a positive integer up to `2147483647`; `0` and malformed values are rejected with `400`, matching Azure) and echoed on container reads and lists.
- Adds the container replace route (`PUT /dbs/{db}/colls/{coll}`) so SDK `replaceContainer` flows can enable, adjust, or (by omitting the property) switch off TTL. The id and partition key are immutable, as on Azure.
- Documents may override the container default with their own `ttl` property (`-1` opts out, positive values shorten or lengthen expiry).
- Expiry follows the [documented semantics](https://learn.microsoft.com/en-us/azure/cosmos-db/time-to-live), measured from `_ts`: with `defaultTtl` absent nothing expires (per-item `ttl` is inert); `-1` enables TTL with no default; `n` expires items `n` seconds after last modification.

As suggested in the issue, enforcement is read-time filtering plus lazy purge: expired documents immediately vanish from point reads, lists, queries, and transactional batches, no longer block re-creating the same id, and are physically deleted when a read next encounters them. This mirrors Azure's contract that expired items leave query results at once while background deletion timing is unspecified.

## Implementation notes

- The decision matrix lives in a small pure class, `CosmosTtl`, evaluated against an explicit clock; the handler passes `Instant.now()` at its call sites. Validation and expiry rules are therefore unit-testable without sleeping.
- `CosmosHandler` gains two helpers (`liveDoc`, `liveDocuments`) used by every document read path, so the batch, query, list, and point-read routes share one enforcement point.
- Write paths already refresh `_ts` on create/replace/patch/upsert, so the TTL clock resets on modification, matching Azure.

## Validation

- `CosmosTtlTest`: the full container/item TTL matrix from the Azure docs, plus validation edge cases (`0`, negatives, fractionals, strings, `2147483648`), with a fixed clock.
- `CosmosContainerTtlTest` (`@QuarkusTest`): the HTTP contract for create/replace (round-trip, disable, invalid values, id/partition-key guards) and one end-to-end expiry test covering reads, lists, queries, batches, and id reuse.
- `CosmosCompatibilityTest#containerTtl`: the real `azure-cosmos` Java SDK against a live emulator run with TLS: `setDefaultTimeToLiveInSeconds` round-trips through create/read/replace, an item using the default expires, and a `ttl = -1` item survives. Verified locally (all 10 tests in the class pass); it will also run in the compat matrix here.
- Full main-module suite: 330 tests, 0 failures.

## Note on #129

This PR and #129 both add a `PUT /dbs/{db}/colls/{coll}` route; the `replaceContainer` here deliberately mirrors the id and partition-key guards from that PR so the merge conflict is mechanical whichever lands second. Happy to rebase this one if #129 goes in first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **cosmos:** container `DefaultTimeToLive` (TTL) support. `defaultTtl` is accepted and validated on container create and replace (`PUT /dbs/{db}/colls/{coll}`), echoed on container reads, and enforced with Azure's semantics measured from `_ts`: absent = TTL off (per-item `ttl` inert), `-1` = TTL on with no default expiry, `n` = expire after `n` seconds, with per-document `ttl` overrides (`-1` opts out). Expired documents immediately vanish from point reads, lists, queries, and transactional batches, no longer block re-creating the same id, and are purged lazily when encountered — mirroring Azure's contract that expired items leave query results at once while background deletion timing is unspecified ([#126](https://github.com/floci-io/floci-az/issues/126))
+
 ## [0.10.0] - 2026-07-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- **cosmos:** container `DefaultTimeToLive` (TTL) support. `defaultTtl` is accepted and validated on container create and replace (`PUT /dbs/{db}/colls/{coll}`), echoed on container reads, and enforced with Azure's semantics measured from `_ts`: absent = TTL off (per-item `ttl` inert), `-1` = TTL on with no default expiry, `n` = expire after `n` seconds, with per-document `ttl` overrides (`-1` opts out). Expired documents immediately vanish from point reads, lists, queries, and transactional batches, no longer block re-creating the same id, and are purged lazily when encountered — mirroring Azure's contract that expired items leave query results at once while background deletion timing is unspecified ([#126](https://github.com/floci-io/floci-az/issues/126))
-
 ## [0.10.0] - 2026-07-31
 
 ### Added
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
@@ -427,4 +427,52 @@ class CosmosCompatibilityTest {
 
         db.delete();
     }
+
+    // --- Time to live ---
+
+    @Test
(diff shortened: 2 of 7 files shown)


hectorvent:
@Ali-Shaikh Thanks for the PR. 
Please resolve conflicts when you have a change.

Ali-Shaikh:
Rebased onto main (0.10.0) and resolved the overlap with #129: container create/replace now validate both `indexingPolicy` and `defaultTtl`, and the composite-index ORDER BY check runs over TTL-filtered documents. Full suite and the Java SDK compat class are green locally.

Ali-Shaikh:
Dropped the CHANGELOG.md hunk in 96a0a3e, same reason as #135: main made the file automation-owned after this branch forked. File matches main again, all checks green.

hectorvent:
Thank you again, and this is ready from my side.

I re-grounded it against the Java SDK this time rather than the Python one, since Java is the primary compat target here. `Constants.DEFAULT_TTL` confirms the `defaultTtl` wire key, and `setDefaultTimeToLiveInSeconds` documents the three states in full. Your matrix matches all of them, including the subtle one: with the container TTL unset, an item level `ttl` is inert rather than taking effect. The `-1` cases behave correctly in both directions too, defaulting to never expire while still letting an item override.

I ran the tests locally as well, 21 passing across `CosmosTtlTest` and `CosmosContainerTtlTest`.

**(follow-up, separate PR)** The item `ttl` itself is not validated, so a document with `"ttl": 0` is quietly treated as never expiring where Azure rejects 0. Not reachable through the SDK, so not worth holding this.

No blockers from my side.

hectorvent (approved):
TTL matrix verified against the Java SDK's documented semantics and the tests run clean locally (21/0). Approving.

PR #142, opened 2026-07-19
fix(tls): emit forward-slash certificate paths so TLS starts on Windows

Ali-Shaikh:
## Summary

Starting the emulator with `FLOCI_AZ_TLS_ENABLED=true` fails on Windows hosts before this fix:

```
java.lang.RuntimeException: Failed to start quarkus
...
Caused by: java.nio.file.NoSuchFileException: D:Devfloci-az.datatlsfloci-az-selfsigned.crt
```

## Root cause

`TlsConfigSource` puts `certFile.toAbsolutePath().toString()` into `quarkus.http.ssl.certificate.files` / `key-files`. On Windows that value contains backslashes (`D:\Dev\floci-az\.\data\tls\floci-az-selfsigned.crt`), and SmallRye Config treats backslashes in property values as escape characters, so by the time the Quarkus HTTP server reads the property every separator has been stripped. Certificate generation itself succeeds; only the config round-trip corrupts the path. User-provided `FLOCI_AZ_TLS_CERT_PATH`/`FLOCI_AZ_TLS_KEY_PATH` values with backslashes were affected the same way.

## Fix

Both emitted paths now go through a small `toConfigPath` helper that swaps separators to forward slashes, but only when `File.separatorChar` is a backslash: the Windows file APIs accept forward slashes, while on unix platforms a backslash is a legal filename character and must pass through untouched. Generated and user-provided paths are both covered, and the behaviour on Linux (including all CI and Docker deployments) is unchanged.

## Reproduction

On a Windows host:

```
./mvnw package -DskipTests
FLOCI_AZ_TLS_ENABLED=true java -jar target/quarkus-app/quarkus-run.jar
```

## Validation

- New `TlsConfigSourceConfigPathTest`: unit coverage of the separator conversion for both platform modes (explicit separator parameter, no OS assumptions), plus assertions that the emitted `quarkus.http.ssl.*` values contain no backslashes and still resolve to existing files, for generated and user-provided certificates.
- Existing `TlsConfigSource*` suites: 24/24 pass.
- Full main-module suite: 313 tests, 0 failures.
- End-to-end on Windows 11: the repro above (clean `data/` directory) now generates the certificate, starts in about a second, and serves both `http://localhost:4577` and `https://localhost:4577` (200 on both, checked against the Cosmos endpoint).

Found while verifying #126 with the Java SDK compatibility suite, which needs the TLS gateway on a Windows dev machine.


hectorvent (approved):
Thanks @Ali-Shaikh,
Solid fix

Issue #120, opened 2026-07-13
[FEAT] Entra ID phase 2: auth-code+PKCE grant + minimal group-membership Graph API


recchia:
## Service

Entra ID / Microsoft Graph

## API Action / Feature

Authorization Code + PKCE grant (interactive) and a narrow slice of Microsoft Graph for group membership:

- `GET /{tenant}/oauth2/v2.0/authorize` — for local dev, could skip real interactive consent (auto-approve against a seeded dev app/user, or a minimal login form) and redirect back with a `code`
- `POST /{tenant}/oauth2/v2.0/token` — accept `authorization_code` + `code_verifier` alongside the existing `client_credentials`/`password` grants
- `POST /v1.0/users/{id}/getMemberGroups` — transitive group membership lookup
- `POST /v1.0/groups/{id}/members/$ref` — add a member to a group
- `DELETE /v1.0/groups/{id}/members/{id}/$ref` — remove a member from a group

Would also need a minimal seed/admin surface to define users, groups, and initial memberships for local dev (could piggyback on the existing dev-tenant seeding pattern from phase 1).

## Azure Documentation

- https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow
- https://learn.microsoft.com/en-us/graph/api/user-getmembergroups
- https://learn.microsoft.com/en-us/graph/api/group-post-members
- https://learn.microsoft.com/en-us/graph/api/group-delete-members

## Why is this needed?

We're using floci-az for local dev and were excited to see Entra ID phase 1 (#23, #73) land — signed JWTs + client_credentials/ROPC cover our backend service-to-service auth entirely offline now. The two things phase 1 explicitly deferred — interactive auth flows and Graph API — are still forcing us out to real Entra locally.

We traced this from a real app using `@azure/msal-react` + Microsoft Graph, and the actual gap is much narrower than "full Graph API": one interactive grant (auth-code+PKCE — no device code needed, since that's the only flow our browser frontend requires) and 3 Graph endpoints for group membership (not full Graph CRUD).

## Are you willing to contribute a PR?

- [x] Yes
- [ ] No

recchia:
Some additional context on why we scoped it this way, plus a couple of questions before we'd start writing code (moved out of the issue body to keep it template-shaped):

[...]

Issue #23, opened 2026-05-20
Add Microsoft Entra ID (Azure AD) support


fredpena:
## Title

Add Microsoft Entra ID (Azure AD) support

## Description

Implement basic Microsoft Entra ID emulation support for floci-az.

The goal is to simplify local authentication and authorization workflows for Azure-based applications.

Suggested capabilities:

* Tenant simulation
* Application registration
* Token generation
* OAuth2/OpenID Connect flows
* Role and permission simulation

## Official Documentation

* https://learn.microsoft.com/en-us/entra/
* https://learn.microsoft.com/en-us/graph/

## Azure REST API Specs

* https://github.com/Azure/azure-rest-api-specs/tree/main/specification/authorization

temaleva:
This is definitely the most important feature. Because otherwise with --connection-string only storage command could be used from azure cli. 

I was able to workaround it using real ENTRA ID endpoint and dummy application. And was even able to perform login. 

Received error later though, trying to use commands because azure cli want all endpoints use https, including management endpoint and cannot workaround it. But at least do not receive please use az login any more.

hectorvent:
Hi @temaleva, 

I am working on it already it will available soon. 

Thanks!

temaleva:
Works for me, thanks! Also using last azure cli without container.

PR #143, opened 2026-07-27
feat: Entra ID phase 2 — auth-code+PKCE grant + Microsoft Graph slice

recchia:
## Summary

Implements #120: the interactive **authorization code + PKCE** grant for Entra ID, plus a narrow **Microsoft Graph** slice for group membership.

- `GET /{tenant}/oauth2/v2.0/authorize` auto-approves against a seeded dev user (no real consent screen, so it stays fully scriptable) and redirects back with a code, honoring `response_mode` (`query`/`fragment`).
- `grant_type=authorization_code` verifies PKCE (`S256`/`plain`), `redirect_uri`, and `client_id`, then issues an access token **and** an ID token (nonce echo, `preferred_username`/`name`/`email`).
- New `graph` service at `/v1.0/...`: `servicePrincipals` discovery (moved from an ad-hoc JAX-RS stub), `POST /v1.0/users/{id}/getMemberGroups`, and `POST`/`DELETE /v1.0/groups/{id}/members/$ref`. Directory data (`User`/`Group`/`GroupMembership`) lives in `EntraStore`, shared with Entra, so a token's `oid` and a Graph lookup resolve to one identity.
- Permissive CORS (`quarkus.http.cors`) so this actually works from a browser SPA (`msal-react`), not just server-to-server clients — confirmed empirically that the preflight is answered before the request reaches `AzureRoutingFilter`.

**Known, explicit limitations** (documented in `docs/services/entra.md`, not silent gaps):
- `refresh_token` grant is out of scope for this PR.
- `getMemberGroups` returns direct membership only — no nested-group transitivity.
- Full Graph CRUD / app-registration management remain in #23.

## Where Graph lives

Per the design discussion on #120: directory data (users/groups/membership) stays in `EntraStore` (the same directory that backs Entra token claims), while the protocol/JSON-envelope logic gets its own `GraphServiceHandler` in a new `services/graph/` package — mirroring how `arm`/`keyvault` are separate handlers sharing one base URL.

## Test plan

- [x] `./mvnw test` — 339/339 passing, including new `EntraStoreTest`, `GraphServiceHandlerTest`, `CorsTest`, and PKCE/nonce round-trip tests in `EntraServiceTest`
- [x] New `compatibility-tests/sdk-test-node/tests/entra.test.ts` — real PKCE codes via `@azure/msal-node`'s `CryptoProvider`, group lookup via the real `@microsoft/microsoft-graph-client` SDK. Verified against a live local emulator (all 4 pass; full node suite 9/9, 79 passed + 1 pre-existing unrelated skip). `make test-entra-node` / `test-entra-node-local` added.
- [ ] CI (native compat matrix) — not run from this environment; the `sdk-test-node` job needs no new env vars, it will pick up the new test file automatically.


hectorvent:
@recchia thanks for PR. 
Please resolve conflicts rebase main and remove the co-author. 

Take a look at the CONTRIBUTORS.md file

recchia:
Rebased onto main (7f58dad, incl. 0.10.0) and resolved the CHANGELOG.md conflicts — the phase 2 Entra/Graph entries now sit under `[Unreleased]` above the already-released `0.10.0` section. Also dropped the `Co-Authored-By` trailer from the last commit per CONTRIBUTING.md's AI-attribution policy. `./mvnw test` still green after the rebase.

hectorvent:
Thank you for rewriting the commits, the attribution is clean now and CI has finally run end to end on this branch: green across CI, Compatibility Tests and Conventional Commits.

Everything I checked on the previous pass still holds at this head. The PKCE S256 verification matches what MSAL actually generates ([PkceGenerator.ts](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/aa48c36/lib/msal-node/src/crypto/PkceGenerator.ts)), the nonce echo is the check MSAL enforces before accepting a response, and the `getMemberGroups` envelope matches the Graph metadata with the direct only behaviour documented as a deviation rather than hidden.

**(follow-up, separate PR)** `EntraStore` now holds mutable membership state but does not implement `Resettable`, so `/_admin/reset` leaves Graph writes behind. Worth a small follow up rather than holding this up.

No blockers from my side.

hectorvent (approved):

PR #146, opened 2026-07-29
fix(blob): return mandatory Get Blob and container response headers

### LINKED ISSUE #145: [BUG] Get Blob omits mandatory response headers, crashing the Azure SDK for C++
## Service

Blob Storage

## Azure API Action

- `Get Blob` (`GET /{container}/{blob}`) and `Get Blob Properties` (`HEAD /{container}/{blob}`)
- `Get Container Properties` (`GET|HEAD /{container}?restype=container`)

## Expected behavior

Per the [Get Blob response headers](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob#response-headers) table, Azure always returns the following on a successful block blob read:

| Header | Spec |
|---|---|
| `x-ms-creation-time` | "Version 2017-11-09 and later. The date/time when the blob was created." |
| `x-ms-server-encrypted` | "Version 2015-12-11 and later. ... `true` if the blob data and application metadata are completely encrypted" |
| `x-ms-lease-status` | "The current lease status of the blob." |
| `x-ms-lease-state` | "Version 2012-02-12 and later. The lease state of the blob." |

The spec's own sample response shows `x-ms-lease-status: unlocked` and `x-ms-lease-state: available`
on a blob with no lease, i.e. these are not conditional on a lease existing.

Relatedly, the spec scopes `Content-Range` to range reads: "Indicates the range of bytes that's
returned in the event that the client requested a subset of the blob by setting the `Range`
request header."

## Actual behavior

floci-az returns none of those four headers. The Azure SDK for C++ reads several of them
unconditionally with `std::map::at()`, which throws `std::out_of_range` when the key is absent, so
the client process aborts rather than surfacing an API error. All line references below are the
current stable release, tag [`azure-storage-blobs_12.18.0`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-storage-blobs_12.18.0/sdk/storage/azure-storage-blobs/src/rest_client.cpp), `sdk/storage/azure-storage-blobs/src/rest_client.cpp`:

| Function | Line | Unguarded read |
|---|---|---|
| `BlobClient::Download` | 3772 | `.at("x-ms-creation-time")` |
| `BlobClient::Download` | 3876 | `.at("x-ms-server-encrypted")` |
| `BlobClient::GetProperties` | 4050 | `.at("x-ms-creation-time")` |
| `BlobClient::GetProperties` | 4158 | `.at("x-ms-server-encrypted")` |
| `BlobContainerClient::GetProperties` | 1480 | `.at("x-ms-lease-state")` |
| `BlobContainerClient::GetProperties` | 1482 | `.at("x-ms-lease-status")` |

```cpp
// BlobClient::Download — no count() guard, unlike every neighbouring header
response.Details.CreatedOn = DateTime::Parse(
    pRawResponse->GetHeaders().at("x-ms-creation-time"),
    Azure::DateTime::DateFormat::Rfc1123);
...
response.Details.IsServerEncrypted
    = pRawResponse->GetHeaders().at("x-ms-server-encrypted") == std::string("true");
```

At the blob level `x-ms-lease-status` / `x-ms-lease-state` are guarded by `count()`, so they are
missing per spec but not fatal. At the container level they are not guarded:

```cpp
// BlobContainerClient::GetProperties — unguarded, unlike x-ms-lease-duration just above
response.LeaseState = Models::LeaseState(pRawResponse->GetHeaders().at("x-ms-lease-state"));
response.LeaseStatus
    = Models::LeaseStatus(pRawResponse->GetHeaders().at("x-ms-lease-status"));
```

floci-az's container response sends `Last-Modified`, `ETag`, `x-ms-has-immutability-policy`, and
`x-ms-has-legal-hold`; four of the six headers this function reads unconditionally, but neither
lease header, so `GetProperties()` on a container throws as well.

The result downstream is a crash loop: any pod whose C++ client downloads a blob, reads blob
properties, or reads container properties terminates and restarts.

The Java, Python, and Node SDKs map these headers to nullable model fields and tolerate their
absence, which is why the existing `sdk-test-java` / `sdk-test-python` / `sdk-test-node` suites are
green and this went unnoticed.

Additionally, `Content-Range` is currently emitted on every `Get Blob` response including full
`200 OK` reads, where the spec expects it only on `206`.

## Reproduction

Header-level, no C++ toolchain needed:

```bash
docker run --rm -p 4577:4577 floci/floci-az:0.9.0   # or: make run

curl -X PUT "http://localhost:4577/devstoreaccount1/c1?restype=container"
curl -X PUT -H "x-ms-blob-type: BlockBlob" --data "hello" \
  http://localhost:4577/devstoreaccount1/c1/b1

# Neither of these prints x-ms-creation-time / x-ms-server-encrypted /
# x-ms-lease-status / x-ms-lease-state; both print Content-Range on a 200.
curl -sD- -o/dev/null    http://localhost:4577/devstoreaccount1/c1/b1   # Get Blob
curl -sD- -o/dev/null -I http://localhost:4577/devstoreaccount1/c1/b1   # Get Blob Properties

# Container properties: no x-ms-lease-state / x-ms-lease-status either.
curl -sD- -o/dev/null "http://localhost:4577/devstoreaccount1/c1?restype=container"
```

With the C++ SDK, `blobClient.Download()`, `blobClient.GetProperties()`, or
`containerClient.GetProperties()` against the same endpoint throws `std::out_of_range` from the
response deserializer.

## Environment

- Floci-az version / image tag: `0.9.0`, also present on current `main` (8b21695). The `getBlob`
  header block is unchanged since the initial commit, so every release is affected.
- Java SDK version (if applicable): n/a; reproduced with the Azure SDK for C++
  `azure-storage-blobs` 12.18.0
- How you're running Floci-az: Docker

mcarvalho-singlestore:
## Summary

`Get Blob`, `Get Blob Properties`, and `Get Container Properties` omitted response headers that Azure always returns, which crashes strict SDK clients. The Azure SDK for C++ reads `x-ms-creation-time`, `x-ms-server-encrypted` (blob) and `x-ms-lease-state`, `x-ms-lease-status` (container) with unguarded `std::map::at()`, so their absence threw `std::out_of_range` inside the response deserializer and took down the calling process, which resulted in a pod crash loop downstream, not a recoverable API error.

This adds the missing headers, records a real creation time, and scopes `Content-Range` to range requests as the spec requires.

Closes #145

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

**Incorrect behavior:** `getBlob` (serving both `GET` and `HEAD`) returned no `x-ms-creation-time`, `x-ms-server-encrypted`, `x-ms-lease-status`, or `x-ms-lease-state`, and sent `Content-Range` on full `200 OK` downloads. `getContainer` returned no `x-ms-lease-state` or `x-ms-lease-status`.

**Contract:** the [Get Blob response headers](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob#response-headers) table lists all four as returned for a block blob read; the spec's sample response shows `x-ms-lease-status: unlocked` / `x-ms-lease-state: available` on an unleased blob. The same table scopes `Content-Range` to "the event that the client requested a subset of the blob by setting the `Range` request header."

**Why it crashes rather than degrades:** `sdk/storage/azure-storage-blobs/src/rest_client.cpp` at tag [`azure-storage-blobs_12.18.0`](https://github.com/Azure/azure-sdk-for-cpp/blob/azure-storage-blobs_12.18.0/sdk/storage/azure-storage-blobs/src/rest_client.cpp) (current stable):

| Function | Line | Unguarded read |
|---|---|---|
| `BlobClient::Download` | 3772 | `.at("x-ms-creation-time")` |
| `BlobClient::Download` | 3876 | `.at("x-ms-server-encrypted")` |
| `BlobClient::GetProperties` | 4050 | `.at("x-ms-creation-time")` |
| `BlobClient::GetProperties` | 4158 | `.at("x-ms-server-encrypted")` |
| `BlobContainerClient::GetProperties` | 1480 | `.at("x-ms-lease-state")` |
| `BlobContainerClient::GetProperties` | 1482 | `.at("x-ms-lease-status")` |

Every neighbouring header in those deserializers is wrapped in `if (...count(...) != 0)`; these are not. At the blob level the two lease headers *are* guarded, so they are added there for spec conformance rather than to fix a crash, but `BlobContainerClient::GetProperties` reads them unguarded, and floci-az sent only four of the six headers that function requires (`Last-Modified`, `ETag`, `x-ms-has-immutability-policy`, `x-ms-has-legal-hold`).

Java / Python / Node map these to nullable model fields, so all three existing compat suites passed against the broken behavior, that gap is why this reached a downstream consumer.

### Notes on the implementation

- **Creation time is persisted, not derived.** Deriving `x-ms-creation-time` from `lastModified` would drift: `setBlobMetadata` and blob overwrite both bump `lastModified`, so a blob's reported creation time would move forward on every metadata update. It is now stamped in the blob's `metadata` map on write, the same mechanism already used for `BlobType`, `Content-Type`, `Name`, and `CommittedBlocks` and preserved on overwrite. `StoredObject` is shared by every service and was deliberately left untouched. Blobs already on disk under the `persistent` / `hybrid` / `wal` backends have no stored value and fall back to `lastModified`.
- **Leases are still not modelled.** The lease headers report the fixed values of an unleased blob or container; there is no way to acquire a lease. `x-ms-server-encrypted: true` mirrors the `x-ms-request-server-encrypted: true` already returned on upload, though nothing is encrypted. Both are now documented as intentional deviations in `docs/services/blob.md` so they are not mistaken for real lease or encryption support.

## Checklist

### Tests

`BlobServiceTest` gains seven cases: the four headers on `GET` and on `HEAD`, the container lease headers on `GET` and `HEAD` `?restype=container`, `Content-Range` absent on `200` and present on `206`, creation time surviving both a metadata update and an overwrite, and creation time on the staged-block `PutBlockList` path. `BlobCompatibilityTest` gains an SDK-level assertion via `blob.getProperties()` (`getCreationTime`, `getLeaseStatus`, `getLeaseState`, `isServerEncrypted`).


hectorvent (approved):
Thanks @mcarvalho-singlestore

PR #148, opened 2026-07-29
fix(email): match the ACS operation contract so the SDKs can complete a send

### LINKED ISSUE #147: [BUG] Email: azure-communication-email Java SDK cannot complete a send — "error": {} fails ErrorDetail deserialization, and "resourceLocation" sends the LRO poller to a 501
The ACS Email data plane is unusable from the `azure-communication-email` **Java** SDK: every
`beginSend(...)` fails, for two independent reasons, plus a third correctness gap. All three come down to
the operation-status payload not matching what the live service sends.

Every emulator-vs-ACS difference below was **verified by signed curl against a live ACS resource**, not
inferred from the spec — and checked on `2023-03-31` (the version floci-az itself defaults to), on
`2024-07-01-preview`, and on `2025-09-01`. **The live response is byte-identical on all three**, so this
is not an api-version mismatch.

Reproduced on **`0.9.0`** and **`nightly-07292026`** — identical on both.

### Environment

- `floci/floci-az:0.9.0` and `floci/floci-az:nightly` (`nightly-07292026`)
- `com.azure:azure-communication-email:1.1.4`, `azure-communication-common:1.4.4`, `azure-core:1.58.1`
- Java 25, `azure-core-http-netty` transport
- Source permalinks below pin `8b21695` (`main` at time of writing)

### Emulator vs. the live service

Emulator:

```console
$ curl -s -X POST "http://localhost:4577/emails:send?api-version=2023-03-31" \
    -H "Content-Type: application/json" -H "Operation-Id: 11111111-2222-4333-a444-555555555555" \
    -d '{"senderAddress":"a@b.c","content":{"subject":"probe"},"recipients":{"to":[{"address":"d@e.f"}]}}'
{"status":"Running","id":"e833f6df-bffe-4d72-8731-5c487a6cc344","error":{}}

$ curl -s "http://localhost:4577/emails/operations/e833f6df-bffe-4d72-8731-5c487a6cc344?api-version=2023-03-31"
{"id":"e833f6df-...","status":"Succeeded","resourceLocation":"50f50f3f-4993-4f9c-884b-aeed18c3d27c","error":{}}
```

Live ACS, HMAC-signed, with `Operation-Id: 440521b4-fa3e-4f6e-afdd-19cf47a5582e` sent:

```console
# POST /emails:send  →  202  (plus a Retry-After: 3 response header)
{"id":"440521b4-fa3e-4f6e-afdd-19cf47a5582e","status":"Running","error":null}

# GET /emails/operations/440521b4-...  →  200
{"id":"440521b4-fa3e-4f6e-afdd-19cf47a5582e","status":"Succeeded","error":null}
```

Three differences: `error` is **`null`, not `{}`**; there is **no `resourceLocation`**; and the client's
**`Operation-Id` is echoed back as `id`**.

---

### Bug 1 — `"error": {}` fails deserialization

https://github.com/floci-io/floci-az/blob/8b21695c1ce29e38fed2a82643e0c897d960b979/src/main/java/io/floci/az/services/email/EmailHandler.java#L170-L178

and the same value on the status path:

https://github.com/floci-io/floci-az/blob/8b21695c1ce29e38fed2a82643e0c897d960b979/src/main/java/io/floci/az/services/email/EmailHandler.java#L203

`EmailSendResult.fromJson` branches on the key being *present*, not on it being meaningful:

```java
} else if ("error".equals(fieldName)) {
    error = ErrorDetail.fromJson(reader);
```

`ErrorDetail` declares `code` and `message` **required**, so an empty object throws:

```
java.io.IOException: java.lang.IllegalStateException: Missing required properties: code, message
```

The live service's `"error": null` goes through the same reader cleanly — a JSON null yields a null
`ErrorDetail`. For a code-generated client, `null` and "present but empty object" are not equivalent.

**Suggested fix:** emit `null`. Worth noting `Map.of("error", null)` throws `NullPointerException`, which
is presumably how the empty map got there in the first place — so the 202 needs a null-tolerant map:

```java
// handleSendEmail — Map.of() cannot hold a null value
Map<String, Object> entity = new LinkedHashMap<>();
entity.put("id", operationId);
entity.put("status", "Running");
entity.put("error", null);

// handleGetOperationStatus — already a LinkedHashMap, so just:
result.put("error", null);
```

### Bug 2 — `resourceLocation` sends the LRO poller to a 501

https://github.com/floci-io/floci-az/blob/8b21695c1ce29e38fed2a82643e0c897d960b979/src/main/java/io/floci/az/services/email/EmailHandler.java#L200-L202

With bug 1 worked around, the send still fails. `resourceLocation` is a **generic** Azure long-running-operation
field, so `azure-core`'s `OperationResourcePollingStrategy.getResult` takes it as the URL of the final
resource:

```java
String finalGetUrl = pollingContext.getData(PollingConstants.RESOURCE_LOCATION);
if (finalGetUrl == null) { ... }   // ← only then does it use the poll body
```

The value is a bare `messageId` UUID, so the SDK issues `GET {endpoint}/<uuid>`, which the emulator
answers with a Storage-shaped XML error, which then fails JSON parsing:

```console
$ curl -s -i http://localhost:4577/50f50f3f-4993-4f9c-884b-aeed18c3d27c
HTTP/1.1 501 Not Implemented
Content-Type: application/xml
x-ms-error-code: NotImplemented

<?xml version='1.0' encoding='UTF-8'?><Error><Code>NotImplemented</Code>...
```

ACS's operation-status response has no `resourceLocation` on any api-version, as shown above. With the
field absent, `getResult` falls through to deserializing the poll body it already holds — the correct
result, and the path the SDK takes in production.

The docs list `resourceLocation` under the operation-status reply, so this may read as intended; the point
is that including it breaks generic Azure LRO pollers, and there is no resource behind the URL it implies.

**Suggested fix:** don't emit it. `messageId` is already exposed via `GET /emailMessages/{operationId}`,
which is a better home for it.

### Bug 3 — `Operation-Id` is ignored

https://github.com/floci-io/floci-az/blob/8b21695c1ce29e38fed2a82643e0c897d960b979/src/main/java/io/floci/az/services/email/EmailHandler.java#L146

`handleSendEmail` always mints its own id and ignores the client's `Operation-Id` request header. The live
service echoes it back as `id`, as shown above.

That header is how a caller reconciles a send it isn't sure completed: mint an id, send it, and after a
crash call `getSendResult(myOperationId)` to learn whether the message actually went. Against the emulator
that lookup always 404s, because the id the caller holds was never the id the emulator used — so
crash-recovery paths can't be tested.

**Suggested fix:** honour the header, falling back to a fresh UUID when absent.

### Minor — no `Retry-After` on the 202

Live ACS returns `Retry-After: 3` alongside the `202`; the emulator returns none and completes the
operation instantly. Not a blocker (pollers fall back to a default interval), but poll-backoff behaviour
can't be exercised, which matters when load-testing against the emulator.

---

### Why this hasn't been caught

Neither blocker surfaces as an ACS `ErrorResponseException` — both look like transport/parse errors — so a
caller with retry logic sees an endlessly retrying send that never actually fails, with no useful
diagnostic. That took a while to track down.

The `compatibility-tests/` suites (`sdk-test-java`, `sdk-test-node`, `sdk-test-python`, Terraform,
OpenTofu, `az` CLI) don't cover email, and there's no `EmailHandlerTest`. The documented quickstart uses
the Python SDK, whose poller tolerates both shapes. A Java suite doing
`beginSend(...).waitForCompletion()` + `getFinalResult()` would catch bugs 1 and 2 in a single assertion,
and asserting the returned `id` catches bug 3.

Happy to open a PR for the handler changes and/or the compat suite if that's useful — everything else in
the emulator worked first try, including TLS with the generated certificate, protocol sniffing on the
shared port, the `Operation-Location` envelope, and the inspection mailbox.

jogerj:
Fixes #147.

The `azure-communication-email` **Java** SDK could not complete a `beginSend`. Four deviations, all in the
two data-plane responses:

| | Was | ACS sends | Effect |
|---|---|---|---|
| 1 | `"error": {}` | `"error": null` | generated `ErrorDetail` requires `code`+`message` → `Missing required properties: code, message` |
| 2 | `"resourceLocation": "<uuid>"` | *(no such field)* | generic LRO field, so `OperationResourcePollingStrategy` GETs `{endpoint}/<uuid>` → `501` in Storage's XML shape → JSON parse failure |
| 3 | `Operation-Id` ignored | caller's id adopted | callers can't reconcile a send whose response they never saw |
| 4 | `Operation-Location` on `base-url`'s scheme | follows the request | an HTTPS client gets an `http://` poll URL its HMAC credential won't sign |

Plus `Retry-After: 3` on the `202`, as ACS does.

1 and 2 each break every send on their own, and neither surfaces as an `ErrorResponseException` — both
arrive as parse failures — so a caller with retry logic sees a send retry forever without ever failing. 2 is
masked by 1, and 4 by both, so they only appear one at a time.

### Where the expected shapes came from

Per `AGENTS.md` ("Do not invent protocol behavior… test against a real Azure endpoint"), signed curl against
a live ACS resource — identical on `2023-03-31` (this emulator's own default), `2024-07-01-preview` and
`2025-09-01`, so none of it is an api-version difference:

```console
# POST /emails:send  →  202, Retry-After: 3, Operation-Id echoed back
{"id":"440521b4-fa3e-4f6e-afdd-19cf47a5582e","status":"Running","error":null}

# GET /emails/operations/440521b4-...  →  200
{"id":"440521b4-fa3e-4f6e-afdd-19cf47a5582e","status":"Succeeded","error":null}
```

### Notes

- `Map.of` rejects null values, which is presumably why `Map.of()` stood in for "no error" — the `202` now
  builds a `LinkedHashMap`.
- Bug 4 uses `EmulatorConfig.baseUrlHttps()`, which already existed for this and had no callers.
  `RequestUrls.resolveBaseUrl` looks like the natural fit but isn't usable: ACS's
  `HmacAuthenticationPolicy` overwrites `Host` with `url.getHost()`, dropping the port, so a
  Host-derived URL sends the poller to `:443`. `TlsProxyServer` serves both schemes on one port, so only
  the scheme needs to vary.
- ARM, the mailbox and routing are untouched.

### Tests

The service had no test, and no `compatibility-tests/` suite covered it — which is why this shipped.

- **`EmailHandlerTest`** — `@QuarkusTest` + RestAssured, 7 cases over the response shapes.
- **`EmailCompatibilityTest`** — real SDK through `beginSend` → `waitForCompletion` → `getFinalResult`,
  which catches 1, 2 and 4 in one assertion, plus `Operation-Id` adoption. Needs
  `FLOCI_AZ_TLS_ENABLED=true` (already set by `docker-compose.yml` and the `Makefile` compat target) since
  the ACS credential only signs `https`; skips itself otherwise.

|  | `EmailHandlerTest` | `EmailCompatibilityTest` |
|---|---|---|
| before | 3 of 7 fail, one per bug | 2/2 error: `Missing required properties: code, message` |
| after | 7/7 | 2/2 |

Full suite: **316 tests, 0 failures**. The compat suite was run against a `Dockerfile.jvm-package` image
built from this branch, and against published `0.9.0` to confirm it fails there.

### Out of scope

APIM, Event Grid, Functions, Monitor and VM also build absolute URLs from `config.effectiveBaseUrl()` and
carry the same scheme assumption. Left alone: those advertise endpoints for a later, separate call, whereas
`Operation-Location` is followed immediately by the same client in the same operation. Happy to generalise
it into a shared helper in a follow-up if you'd prefer that to a per-handler fix.

### DIFF
--- a/compatibility-tests/sdk-test-java/pom.xml
+++ b/compatibility-tests/sdk-test-java/pom.xml
@@ -74,6 +74,10 @@
             <artifactId>azure-messaging-eventgrid</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-communication-email</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
             <version>${junit.version}</version>
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/EmailCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/EmailCompatibilityTest.java
@@ -1,0 +1,101 @@
+package io.floci.az.compat;
+
+import com.azure.communication.email.EmailClient;
+import com.azure.communication.email.EmailClientBuilder;
+import com.azure.communication.email.models.EmailAddress;
+import com.azure.communication.email.models.EmailMessage;
+import com.azure.communication.email.models.EmailSendResult;
+import com.azure.communication.email.models.EmailSendStatus;
+import com.azure.core.http.HttpHeaderName;
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
+import com.azure.core.util.polling.SyncPoller;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import reactor.core.publisher.Mono;
+
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("Communication Services Email Compatibility")
+class EmailCompatibilityTest {
+
+    private static final String SENDER = "DoNotReply@example.com";
+    private static final String RECIPIENT = "dev@example.com";
+
+    private EmailClient client;
+
+    @BeforeAll
+    void setUp() throws Exception {
+        EmulatorConfig.assumeEmulatorRunning();
+        EmulatorConfig.installEmulatorTlsCert();
+        Assumptions.assumeTrue(
+                System.getProperty("javax.net.ssl.trustStore") != null,
+                "requires FLOCI_AZ_TLS_ENABLED=true — the ACS credential only signs https URLs");
+        client = buildClient(null);
+    }
+
+    @Test
+    @DisplayName("beginSend completes through the SDK's long-running-operation poller")
+    void beginSendCompletesThroughTheSdkPoller() {
+        SyncPoller<EmailSendResult, EmailSendResult> poller = client.beginSend(message("SDK poller"));
+        poller.waitForCompletion();
+
+        EmailSendResult result = poller.getFinalResult();
+        assertNotNull(result);
+        assertEquals(EmailSendStatus.SUCCEEDED, result.getStatus());
+        assertNotNull(result.getId());
+        assertNull(result.getError());
+    }
+
+    @Test
+    @DisplayName("a caller-supplied Operation-Id is adopted as the operation id")
+    void callerSuppliedOperationIdIsAdopted() {
+        String operationId = UUID.randomUUID().toString();
+
+        SyncPoller<EmailSendResult, EmailSendResult> poller =
+                buildClient(operationId).beginSend(message("Operation-Id echo"));
+        poller.waitForCompletion();
+
+        assertEquals(operationId, poller.getFinalResult().getId());
+    }
+
+    private EmailClient buildClient(String operationId) {
+        String endpoint = EmulatorConfig.httpBase().replace("http://", "https://");
+        EmailClientBuilder builder = new EmailClientBuilder()
+                .connectionString("endpoint=" + endpoint + "/;accesskey=" + EmulatorConfig.DEV_KEY);
+        if (operationId != null) {
+            builder.addPolicy(new StaticOperationIdPolicy(operationId));
+        }
+        return builder.buildClient();
+    }
+
+    private static EmailMessage message(String subject) {
+        return new EmailMessage()
+                .setSenderAddress(SENDER)
+                .setSubject(subject)
+                .setBodyHtml("<p>floci-az compatibility test</p>")
+                .setToRecipients(List.of(new EmailAddress(RECIPIENT)));
+    }
+
+    private record StaticOperationIdPolicy(String operationId) implements HttpPipelinePolicy {
+
+        private static final HttpHeaderName OPERATION_ID = HttpHeaderName.fromString("Operation-Id");
+
+        @Override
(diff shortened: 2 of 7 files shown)


hectorvent (approved):
Thanks @jogerj

PR #150, opened 2026-07-30
feat: add MySQL, MariaDB, and diagnostic settings emulation

hectorvent:
## Summary

Continues #104 by @guilhermefacanha — his commits are preserved with him as author, ported onto the handler-owned routing model (`routes()`/`enabled()`/`Resettable`).

- ARM management-plane emulation for **Azure Database for MySQL (Flexible Server)** (`Microsoft.DBforMySQL/flexibleServers`) and **Azure Database for MariaDB (Single Server)** (`Microsoft.DBforMariaDB/servers`): server CRUD, databases, firewall rules, configurations, checkNameAvailability, plus a `/connect` convenience endpoint — backed by real `mysql:8.0` / `mariadb:10.11` sidecar containers.
- Cross-resource **diagnostic settings** extension in MonitorHandler (`{resourceUri}/providers/microsoft.insights/diagnosticSettings`), claimed in the routeMonitor stage before ARM provider dispatch.
- Follow-up fixes surfaced by the new compat suites: readiness now waits for the MySQL protocol handshake instead of a bare TCP accept (Docker's port proxy accepts long before mysqld listens), and the admin user gets global privileges via an init script, matching a real Azure admin's ability to create databases.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Wire shapes follow the ARM contracts at the api-versions the azurerm Terraform provider pins: `Microsoft.DBforMySQL` FlexibleServers `2023-12-30` and `Microsoft.DBforMariaDB` `2018-06-01` (verified against azure-rest-api-specs and the provider's vendored go-azure-sdk). Validated with JDBC drivers `mysql-connector-j 8.4.0` and `mariadb-java-client 3.4.1` against live sidecar containers.

## Checklist

- [x] `./mvnw test` passes locally (367 tests)
- [x] New or updated integration test added (44 unit tests + 20 Java SDK compatibility tests, `make test-mysql` / `make test-mariadb`)

Refs #104

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
         test-blob test-blob-local test-blob-python test-blob-python-local test-blob-java test-blob-java-local test-blob-node test-blob-node-local \
         test-apim-java \
         test-cosmos test-cosmos-mongo test-cosmos-postgresql test-cosmos-cassandra test-cosmos-gremlin test-cosmos-table test-cosmos-nosql test-cosmos-all \
-        test-sql test-terraform-compat test-opentofu-compat test-azcli test-iac-compat compat-docker test-compat clean
+        test-sql test-mysql test-mariadb test-terraform-compat test-opentofu-compat test-azcli test-iac-compat compat-docker test-compat clean
 
 MVN            = ./mvnw
 PORT           = 4577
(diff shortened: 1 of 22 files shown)


hectorvent:
Addressed all three review findings in 652e415:

1. **Handshake timeout → failed create**: `waitForReady` now throws on timeout, so the create path returns 500 with state rollback instead of reporting `Ready` for a server that never spoke the protocol.
2. **Rehydration after restart**: persisted servers with `containerId=null` are lazily restarted on `/connect` and on PUT/PATCH (new `ensureStarted`, reusing the create-path start locks) — the `loadFromStore` log message is now accurate.
3. **Password rotation**: PUT/PATCH with a new `administratorLoginPassword` now runs `ALTER USER` inside the live container via a new `ContainerLifecycleManager.execInContainer` API. Rotation happens before the state commit, so a failed rotation leaves state matching what the container still accepts. Covered by new compat tests that verify the new password authenticates over JDBC and the old one is rejected (22/22 passing against live containers).

Note: findings 1 and 2 also exist in the PostgreSQL service this pattern was copied from — worth a separate issue.

hectorvent:
cbc8dff addresses the remaining 3/5 blocker: restored servers now rehydrate on ordinary GET as well, not just /connect and PUT/PATCH — `terraform refresh` and SDK pollers read state with a bare GET, so a persisted server comes back `Ready` there after an emulator restart. Rehydration on read is best-effort: if the container cannot be restarted, the GET still succeeds and reports the current state rather than failing the read. Verified: 367/367 unit tests, 22/22 live compat tests.

hectorvent:
Fixed the orphaned-container finding: `startServer` now stops and removes its own container before rethrowing on any start failure (handshake timeout, port resolution, copy). The handler's rollback removes a state entry that never learned the `containerId`, so the manager is the only place that can dispose of it. Verified again: 367/367 unit, 22/22 live compat.

PR #157, opened 2026-07-31
feat(docker): label emulator containers and volumes, centralize the floci-az name prefix

hectorvent:
## Summary

Implements the cross-emulator Docker attribution convention for floci-az: every emulator-created container and volume is labeled `floci=true` (umbrella across the Floci emulators) plus `floci_emulator=floci-az` (per-emulator discriminator), and `floci_namespace` when configured — applied centrally at the container lifecycle choke point, so `docker ps --filter label=floci_emulator=floci-az` and `docker volume prune --filter label=floci_emulator=floci-az` target this emulator alone while `label=floci=true` still matches all of them.

- The `floci-az-` name prefix is now owned by `ContainerStorageHelper` instead of being spelled at eleven call sites — **container names are unchanged**.
- The Functions runtime container is created through the shared `ContainerBuilder`/`ContainerLifecycleManager` layer (it previously issued its own `createContainerCmd` and carried no labels).
- The AKS k3s volume is created explicitly via `ensureVolume` (its first caller) so it is labeled too.
- New optional `floci-az.docker.resource-namespace` (`FLOCI_AZ_DOCKER_RESOURCE_NAMESPACE`) inserts a namespace into child container/volume names (`floci-az-<ns>-…`) and a `floci_namespace` label, for running multiple emulator processes against one Docker daemon.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

n/a for wire protocol — operational Docker hygiene. Verified live: `docker inspect` on an emulator-created container shows `{"floci":"true","floci_emulator":"floci-az"}` and the name is byte-identical to before (`floci-az-pg-…`). Documented caveat: with a resource-namespace set, the Event Hubs broker certificate SAN (`floci-az-artemis`) does not cover the namespaced container name for in-network TLS clients — connect via `localhost` in that setup.

## Checklist

- [x] `./mvnw test` passes locally (567 tests)
- [x] New or updated integration test added — `ContainerLifecycleManagerLabelsTest` + `ContainerStorageHelperTest` (10 cases asserting labels/naming at the choke point, mirroring the floci-gcp reference implementation), plus a live-Docker label verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **docker:** every emulator-created container and volume is now labelled `floci=true`
+  (umbrella across the Floci emulators) and `floci_emulator=floci-az` (per-emulator
+  discriminator), applied centrally in the container lifecycle layer — so
+  `docker ps --filter label=floci_emulator=floci-az` and
+  `docker volume prune --filter label=floci_emulator=floci-az` target this emulator
+  alone, while `label=floci=true` still matches all Floci emulators. The `floci-az-`
+  name prefix is now owned by a single naming helper instead of being spelled at each
+  call site (container names are unchanged), the Functions runtime container is created
+  through the shared container layer (it previously bypassed it and carried no labels),
+  and the AKS k3s volume is created explicitly so it is labelled too. New optional
+  `floci-az.docker.resource-namespace` (`FLOCI_AZ_DOCKER_RESOURCE_NAMESPACE`) inserts a
+  namespace into child container/volume names (`floci-az-<ns>-...`) and a
+  `floci_namespace` label, for running multiple emulator processes against one Docker
+  daemon. Note: with a namespace set, the Event Hubs broker certificate SAN
+  (`floci-az-artemis`) does not cover the namespaced container name for in-network TLS
+  clients — connect via `localhost` in that setup.
+
+### Added
+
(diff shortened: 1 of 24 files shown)


hectorvent:
Addressed the review findings:

1. **MySQL/MariaDB namespace bypass** — real gap (those managers postdate the call-site survey): both now route through `ContainerStorageHelper.dockerName`, so they honor the resource-namespace and stay byte-identical without one.
2. **AKS volume leak on failed start** — `startCluster` now removes the volume on a failed create/start, but **only when that start created it**: a pre-existing volume may hold a previous cluster's k3s state and must survive transient failures.
3. **Pre-existing unlabeled volumes** — documented rather than migrated: Docker volume labels are immutable (there is no relabel API), so retroactively labeling would require destroy-and-recreate. The CHANGELOG now carries the caveat with a manual listing/cleanup command; in practice the AKS volume is deleted on shutdown by default, so the population self-resolves. Same conclusion the floci-gcp port reached.

567/567 tests green.

hectorvent:
Good catch on the post-start window: the failure path now force-removes the container by name **before** the volume, so a container that started but failed inspection no longer keeps running and pinning the volume. The volume removal stays gated on this start having created it.

On legacy unlabeled volumes: holding the documented-caveat position deliberately. Docker has no relabel API, so the only code path is destroy-and-recreate of user volumes — trading a label gap (they remain attributable by the `floci-az-` name prefix) for data destruction. Same trade-off the floci-gcp port settled on.

PR #158, opened 2026-07-31
feat: enforce scoped ADLS SAS authorization

snazy:
## Summary

- Add parsed storage SAS context and deterministic shared user-delegation signing key material.
- Validate user-delegation SAS signatures, resource scope, path scope, and operation permissions before Blob-backed ADLS operations.
- Add Blob and Java DataLake SDK compatibility coverage for accepted and denied SAS behavior.
- Signed user-delegation key-window validation now includes: skt/ske, signed key identity/service, and SAS expiry bounded by key expiry

## Type of change

- [x] New feature (feat:)

## Azure Compatibility

- Matches Java Azure SDK DataLake user-delegation SAS string-to-sign behavior for the compatibility-suite SDK version.
- SDK test source compile verified with repo-local Maven cache.

## Checklist

- [x] New or updated integration test added.
- [x] Commit messages follow Conventional Commits.


hectorvent:
Thank you for taking this from parsing to actual enforcement, and for keeping the new types in `core/auth` alongside the existing verifiers rather than starting a parallel path.

Two things, and the first one is our fault. We merged #166 today and it touched `BlobServiceHandler`, which has put this branch into conflict. Only `BlobServiceTest.java` actually conflicts, so it should be a small rebase, and I am sorry for the churn.

The second is that I have not finished checking the SAS field semantics against Azurite yet. Given this flips a permissive path into an enforcing one I would rather not sign off from a skim. The pieces I want to walk properly are the signed permission to operation mapping, the `sr` scoping including the directory depth case, and the user delegation key fields.

Nothing looks wrong so far. I will come back with specifics once you have rebased.

hectorvent:
Thank you for this, and sorry it sat with me so long. A change that turns a permissive path into an enforcing one deserved a proper pass rather than a skim, and I have now done it.

The signing is right. I compared the string-to-sign field by field against Azurite and your ordering matches its `sv >= 2020-12-06` layout exactly, including `ses` sitting between the blob version timestamp placeholder and the response header overrides, which is the easy one to get wrong. Deriving the canonical name from the first `sdd` segments so that containment is enforced by the signature itself, rather than by a prefix comparison, is also the right shape.

Three things worth a follow up issue rather than holding this: the signing uses the 2020-12-06 layout regardless of `sv`, the `a` permission is not mapped so an append only SAS would be refused, and `sr=bs` is rejected. All three fail closed, which is the safe direction.

No blockers from my side.

snazy:
> Thank you for this, and sorry it sat with me so long. A change that turns a permissive path into an enforcing one deserved a proper pass rather than a skim, and I have now done it.
> 
> The signing is right. I compared the string-to-sign field by field against Azurite and your ordering matches its `sv >= 2020-12-06` layout exactly, including `ses` sitting between the blob version timestamp placeholder and the response header overrides, which is the easy one to get wrong. Deriving the canonical name from the first `sdd` segments so that containment is enforced by the signature itself, rather than by a prefix comparison, is also the right shape.
> 
> Three things worth a follow up issue rather than holding this: the signing uses the 2020-12-06 layout regardless of `sv`, the `a` permission is not mapped so an append only SAS would be refused, and `sr=bs` is rejected. All three fail closed, which is the safe direction.
> 
> No blockers from my side.

Thanks!
I'll prepare commits for follow-up PRs (on top of a change in the PR stack #121).

PR #166, opened 2026-07-31
fix(blob): persist upload HTTP properties

### LINKED ISSUE #163: [BUG] Blob: Put Blob ignores x-ms-blob-content-* properties
## Service

Blob Storage

## Azure API Action

Persist Blob HTTP properties supplied through `x-ms-blob-content-*` upload headers, especially `x-ms-blob-content-type` and `x-ms-blob-cache-control`.

Azure documentation: https://learn.microsoft.com/rest/api/storageservices/put-blob#request-headers-all-blob-types

## Expected behavior

A `Put Blob` request containing:

```http
x-ms-blob-type: BlockBlob
x-ms-blob-content-type: image/png
x-ms-blob-cache-control: public, max-age=31536000
Content-Type: application/octet-stream
```

stores Blob properties `Content-Type: image/png` and `Cache-Control: public, max-age=31536000`. HEAD / Get Properties and Get Blob return them.

## Actual behavior

0.10.0 stores the HTTP request entity's `Content-Type` and ignores the `x-ms-blob-content-*` property headers.

Live result:

```text
Put Blob: 201
HEAD:     200
Content-Type: application/octet-stream
Cache-Control: <absent>
```

`BlobServiceHandler.putBlob` reads `HttpHeaders.CONTENT_TYPE` but does not read `x-ms-blob-content-type` or `x-ms-blob-cache-control`.

## Why this matters

Azure Storage SDKs use these headers for `BlobHttpHeaders`. Direct-upload workflows commonly send `x-ms-blob-content-type`, then read Blob properties to validate uploaded media. Floci-AZ reports such uploads as `application/octet-stream`, breaking completion and media-type validation. Server-side uploads also depend on cache-control persistence.

Suggested acceptance:

- Honor Azure precedence for `x-ms-blob-content-type`, `x-ms-blob-cache-control`, content disposition, encoding, language, and MD5 where supported.
- Return stored properties from HEAD and GET.
- Add a real SDK compatibility case using `BlobHttpHeaders`.

## Environment

- SDK: `Azure.Storage.Blobs` 12.29.1
- Docker Desktop on Windows

thomhurst:
## Summary

- persist Azure Blob HTTP properties supplied through `x-ms-blob-content-*` during Put Blob
- prefer `x-ms-blob-content-type` over request entity `Content-Type`, matching Azure
- return stored content type, cache control, disposition, encoding, language, and MD5 from HEAD/GET
- add REST regression coverage and Azure Storage Java SDK compatibility coverage

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

0.10.0 stored the request entity `Content-Type` and discarded Blob HTTP property headers. This now follows Put Blob header semantics used by `BlobHttpHeaders`; stored properties round-trip through Get Blob / Get Properties.

## Checklist

- [x] `./mvnw test` passes locally (554 tests)

Closes #163

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/BlobCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/BlobCompatibilityTest.java
@@ -6,6 +6,7 @@ import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.blob.models.BlobErrorCode;
+import com.azure.storage.blob.models.BlobHttpHeaders;
 import com.azure.storage.blob.models.BlobItem;
 import com.azure.storage.blob.models.BlobProperties;
 import com.azure.storage.blob.models.BlobRange;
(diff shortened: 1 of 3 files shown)


hectorvent:
Thank you, this is a clean fix and it gets the fiddly part right.

The six headers you map are exactly the six that Put Blob accepts, and mapping them onto the bare `Content-*` and `Cache-Control` names on the way back out is the important detail: Get Blob Properties returns them under those names rather than the `x-ms-blob-` prefixed ones, so echoing the request names back would have looked correct locally and still confused the SDK.

Suppressing `Content-MD5` on range reads is a nice touch too, that matches how real Azure treats a ranged GET.

No blockers from my side.

hectorvent (approved):
Header mapping verified against the Put Blob and Get Blob Properties contracts, including the request to response rename. Approving.

PR #167, opened 2026-07-31
fix(cosmos): enforce item ETag conditions

### LINKED ISSUE #160: [BUG] Cosmos DB: item writes ignore ETag preconditions
## Service

Cosmos DB (NoSQL / SQL API)

## Azure API Action

Optimistic concurrency for item replace, delete, patch, and upsert operations using `If-Match` / `If-None-Match`, including per-operation preconditions inside transactional batches.

## Expected behavior

A stale `If-Match` ETag fails with HTTP `412 Precondition Failed`; the item remains unchanged. Transactional-batch item options enforce the same precondition.

Azure documentation: https://learn.microsoft.com/azure/cosmos-db/nosql/database-transactions-optimistic-concurrency#implement-optimistic-concurrency-control-using-etag-and-http-headers

## Actual behavior

0.10.0 ignores the precondition and applies the write. Reproduced against the live image:

```http
PUT /devstoreaccount1-cosmos/dbs/compat-audit/colls/items/docs/one
If-Match: "definitely-stale"
x-ms-documentdb-partitionkey: ["p"]
```

Result: HTTP `200`; document was replaced. `CosmosHandler` does not inspect conditional headers on document write/delete paths, and batch helpers do not inspect per-operation request options.

## Why this matters

Cosmos applications use `ItemRequestOptions.IfMatchEtag` on concurrency-sensitive updates and `TransactionalBatchItemRequestOptions.IfMatchEtag` in batches. Ignoring ETags turns lost-update protection into unconditional last-write-wins locally, so concurrency tests pass while production correctly returns 412.

Suggested acceptance:

- Correct `If-Match` and `If-None-Match` behavior on item replace, delete, patch, and upsert.
- Correct 412 operation result inside transactional batches.
- Stored `_etag` and response `ETag` stay aligned after successful mutations.
- Compatibility coverage through a real Azure Cosmos SDK, including the .NET SDK if possible.

## Environment

- Reproduced with raw Cosmos REST requests; affected SDK is `Microsoft.Azure.Cosmos` 3.61.0
- Docker Desktop on Windows

thomhurst:
## Summary

- Enforce `If-Match` and `If-None-Match` on point create, upsert, replace, patch, and delete paths.
- Enforce per-operation ETag conditions for transactional-batch create, upsert, replace, patch, and delete.
- Commit successful transactional batches through `StorageBackend.applyBatch`; failed batches leave storage unchanged.
- Return Azure-compatible `412 PreconditionFailed` and `424 FailedDependency` responses.
- Add HTTP, storage, and Azure Cosmos Java SDK regression coverage.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Breaking change
- [ ] Docs / chore

## Validation

- `.\mvnw.cmd test "-Dtest=CosmosItemConditionTest,CosmosTransactionalBatchTest,StorageBatchTest"` — 14 tests passed.
- Existing PR checks passed before the follow-up; checks rerun after this push.

Depends on #169.

Closes #160

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
@@ -247,7 +247,7 @@ class CosmosCompatibilityTest {
     }
 
     @Test
-    @DisplayName("transactional batch: Create/Read/Replace/Delete/Upsert")
+    @DisplayName("transactional batch: CRUD, Patch, and rollback")
     @SuppressWarnings("unchecked")
     void transactionalBatch() {
         String id = dbId();
@@ -289,6 +289,32 @@ class CosmosCompatibilityTest {
                 .getItem().get("v")).intValue());
         assertThrows(CosmosException.class,
                 () -> container.readItem("b3", new PartitionKey("test"), Map.class));
+
+        // Batch 3: Patch is visible to a later operation in the same batch.
+        CosmosBatch batch3 = CosmosBatch.createCosmosBatch(new PartitionKey("test"));
+        batch3.patchItemOperation("b1", CosmosPatchOperations.create()
+                .set("/v", 7)
+                .increment("/counter", 2));
+        batch3.readItemOperation("b1");
+
+        CosmosBatchResponse r3 = container.executeCosmosBatch(batch3);
+        assertEquals(200, r3.getResults().get(0).getStatusCode());
+        assertEquals(200, r3.getResults().get(1).getStatusCode());
+        assertEquals(7, ((Number) container
+                .readItem("b1", new PartitionKey("test"), Map.class)
+                .getItem().get("v")).intValue());
+
+        // Batch 4: a failure rolls back an earlier successful mutation.
+        CosmosBatch batch4 = CosmosBatch.createCosmosBatch(new PartitionKey("test"));
+        batch4.replaceItemOperation("b1", doc("b1", "test", "v", 999));
+        batch4.deleteItemOperation("missing");
+
+        CosmosBatchResponse r4 = container.executeCosmosBatch(batch4);
+        assertEquals(424, r4.getResults().get(0).getStatusCode());
+        assertEquals(404, r4.getResults().get(1).getStatusCode());
+        assertEquals(7, ((Number) container
+                .readItem("b1", new PartitionKey("test"), Map.class)
+                .getItem().get("v")).intValue());
 
         db.delete();
     }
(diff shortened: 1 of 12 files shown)


hectorvent:
Thank you for this, and for putting the batch primitive on `StorageBackend` and implementing it across all four backends rather than special casing it in the Cosmos handler. That is the right seam.

Two things have come up, and both are on our side rather than yours.

First, we merged #143 today and it added a `remove()` method across those same four backends, so this branch now conflicts on `StorageBackend`, `InMemoryStorage` and `PersistentStorage`. That is a real interface level overlap rather than just appended tests, so it needs a little care on the rebase. Apologies for the timing.

Second, this PR appears to contain all of #169: the `StorageBackend` additions are identical and both add the same three test files, under different commit hashes. Could you confirm whether #169 is meant to be superseded by this one? If so I would rather take this and close that, so you are not maintaining it twice.

thomhurst:
Yes, #167 supersedes #169. It contains the complete transactional-batch and storage work from #169 together with the ETag-condition fixes. I have rebased #167 onto current main and closed #169 as included. Thanks for flagging the overlap.

hectorvent:
Thank you for closing #169, that settled the duplication cleanly and this now carries the batch work in one place.

Unfortunately we merged #141 (container TTL) since, which has put this into conflict on `CosmosHandler.java`. That is our doing and I am sorry for the churn.

One thing worth knowing before you rebase, because I do not think it is a take both sides resolution. On main every document read now goes through `liveDoc(store.get(key), defaultTtl)`, so an expired item reads as absent. Your precondition and batch paths read the raw map directly in five places, which was right before TTL existed. If those stay raw after the rebase, an `If-Match` against an expired but not yet purged item would return 412 where a plain read would say 404.

Routing those reads through the same `liveDoc` helper should keep the two features consistent. Happy to look again once it is rebased.

thomhurst:
Rebased onto current main and resolved the `CosmosHandler` overlap by routing both point-mutation and staged-batch reads through `liveDoc`. I also added regression coverage for expired documents under conditional point and transactional batch mutations. The focused Cosmos/storage suite is green (24 tests).

hectorvent:
Thank you, that is exactly it, and routing the batch reads through a single `liveBatchDocument` helper is neater than patching each call site the way I described.

With that in place an expired item is now invisible to the precondition checks in the same way it is to a plain read, so the 412 against something a read would report as 404 cannot happen. The rebase itself came out clean and CI is green.

Thanks also for closing #169 earlier, that made this one much easier to reason about.

No blockers from my side.

hectorvent (approved):
Rebase resolved the conflict and the TTL interaction is handled via a single liveBatchDocument seam, so expired items are absent to the precondition paths too. Approving.

PR #168, opened 2026-07-31
fix(cosmos): evaluate correlated EXISTS queries

### LINKED ISSUE #162: [BUG] Cosmos DB: unsupported correlated EXISTS predicate is silently ignored
## Service

Cosmos DB (NoSQL / SQL API)

## Azure API Action

Correlated `EXISTS` subqueries over nested arrays.

Azure documentation: https://learn.microsoft.com/azure/cosmos-db/nosql/query/subquery#exists-expression

## Expected behavior

This query returns only documents containing a matching nested action:

```sql
SELECT * FROM c
WHERE EXISTS(
    SELECT VALUE action
    FROM action IN c.actions
    WHERE action.resolvesOn.type = @type)
```

If correlated subqueries remain unsupported, Floci-AZ should reject the query with a clear 400 rather than return incorrect data.

## Actual behavior

0.10.0 returns HTTP 200 but silently ignores the unsupported `EXISTS` predicate. A document with `"actions": []` is returned for `@type = "rsvp"`.

This is more dangerous than a documented unsupported-query error: callers receive valid-looking false positives. The current Cosmos docs list nested-array JOIN as unsupported, but do not mention `EXISTS`, and the query engine does not fail closed.

## Why this matters

Applications use this query shape to find documents with matching nested actions. False positives can send unrelated records into downstream processing, and local tests can approve behavior that differs observably from Azure.

Suggested acceptance:

- Implement correlated `EXISTS` + `FROM ... IN` semantics; or first fail such queries explicitly with HTTP 400.
- Add positive and negative SDK compatibility cases, including an empty array and a non-matching nested value.
- Document the behavior in `docs/services/cosmos.md`.

## Environment

- Reproduced with Cosmos REST query request
- SDK: `Microsoft.Azure.Cosmos` 3.61.0
- Docker Desktop on Windows

thomhurst:
## Summary

- evaluate correlated `EXISTS` subqueries over nested document arrays
- bind each array object as the subquery alias and evaluate its nested `WHERE` predicate
- handle empty, non-matching, and matching arrays without false positives
- document correlated `EXISTS` support and add engine plus HTTP regression tests

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

0.10.0 interpreted the comparison inside an unsupported `EXISTS` expression as `null = null`, returning every document. The query engine now follows Cosmos correlated array-subquery semantics for `SELECT VALUE alias FROM alias IN c.array [WHERE ...]`.

## Checklist

- [x] `./mvnw test` passes locally (555 tests)

Closes #162

### DIFF
--- a/docs/services/cosmos.md
+++ b/docs/services/cosmos.md
@@ -9,7 +9,7 @@ Compatible with the `azure-cosmos` SDK (Java, Python, JavaScript, .NET).
 - **Documents** — create, get, replace, delete, list; upsert via `x-ms-documentdb-is-upsert` header
 - **Queries** — in-process SQL engine with full Cosmos DB SQL dialect support:
   - `SELECT *`, `SELECT c.field1, c.field2`, `SELECT VALUE c.field`, `SELECT TOP n`
-  - `WHERE` with `=`, `!=`, `<>`, `>`, `>=`, `<`, `<=`, `IN`, `BETWEEN`, `NOT`, `AND`, `OR`
+  - `WHERE` with `=`, `!=`, `<>`, `>`, `>=`, `<`, `<=`, `IN`, `BETWEEN`, `NOT`, `AND`, `OR`, and correlated `EXISTS` over arrays
   - `WHERE` functions: `IS_DEFINED`, `IS_NULL`, `IS_STRING`, `IS_NUMBER`, `IS_BOOL`, `IS_ARRAY`, `IS_OBJECT`, `CONTAINS`, `STARTSWITH`, `ENDSWITH`, `ARRAY_CONTAINS`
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you for this one, silently ignoring a predicate is the worst kind of query bug because the caller gets a plausible looking result set rather than an error, so it can sit undetected for a long time.

Preserving the alias scope so a correlated subquery actually sees the outer item, while an uncorrelated one does not leak into it, is the right split, and the added tests cover both directions rather than just the happy path. Updating `docs/services/cosmos.md` alongside it is appreciated.

Worth noting this is the one Cosmos change of the current three that stands entirely on its own, it touches only the query engine, so it does not need to queue behind the others.

No blockers from my side.

hectorvent (approved):
Verified against the query semantics and the added tests cover both the correlated and uncorrelated directions. Approving.

PR #169, opened 2026-07-31
fix(cosmos): make transactional batches atomic

### LINKED ISSUE #161: [BUG] Cosmos DB: transactional batch partially commits and rejects Patch
## Service

Cosmos DB (NoSQL / SQL API)

## Azure API Action

Transactional batch all-or-nothing semantics and `Patch` operations.

Azure documentation: https://learn.microsoft.com/azure/cosmos-db/nosql/transactional-batch

## Expected behavior

- All operations in one logical-partition batch commit atomically.
- If one operation fails, earlier operations are rolled back and dependent results report failed dependency.
- `Patch` is accepted as a transactional-batch operation, including `set` and `incr` operations.

## Actual behavior

0.10.0 mutates the store operation-by-operation and never rolls back. It also rejects `Patch` with a per-operation 400.

Live reproduction:

```json
[
  {
    "operationType": "Replace",
    "id": "1",
    "resourceBody": { "id": "1", "pk": "p", "value": 777 }
  },
  { "operationType": "Delete", "id": "missing" }
]
```

Response contains operation statuses `[200, 404]`, but a point read shows item `1` permanently changed to `777`. On Azure the failed batch leaves item `1` unchanged.

A batch containing:

```json
{
  "operationType": "Patch",
  "id": "2",
  "resourceBody": {
    "operations": [{ "op": "set", "path": "/isRead", "value": true }]
  }
}
```

returns operation status 400. `CosmosHandler.executeBatch` currently supports only Create, Upsert, Read, Replace, and Delete.

## Why this matters

Applications use transactional batches for partition-scoped patches and cleanup deletes. Partial commit breaks the atomicity invariant the API is meant to provide; missing batch Patch also prevents normal SDK workflows.

Suggested acceptance:

- Stage batch mutations and commit only if every operation succeeds.
- Return Azure-compatible failed-dependency results after the first failure.
- Support `Patch` with the same operation set as point PATCH.
- Cover rollback and batch Patch through a real SDK compatibility test.

## Environment

- SDK: `Microsoft.Azure.Cosmos` 3.61.0
- Docker Desktop on Windows

thomhurst:
## Summary

- execute transactional-batch operations against a partition-local staged snapshot
- commit staged writes only when every operation succeeds; failed batches leave storage unchanged
- return the failing status for the failed operation and `424 Failed Dependency` for rolled-back/skipped operations
- support transactional-batch Patch using the same `add`, `set`, `replace`, `remove`, `incr`, and `move` implementation as point Patch
- add HTTP rollback/Patch tests and expand Azure Cosmos Java SDK compatibility coverage

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

0.10.0 mutated storage operation-by-operation, so failed batches partially committed, and returned 400 for Patch. Batches now have all-or-nothing commit behavior and Azure-style per-operation results.

## Checklist

- [x] `./mvnw test` passes locally (555 tests)

Closes #161

### DIFF
--- a/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
+++ b/compatibility-tests/sdk-test-java/src/test/java/io/floci/az/compat/CosmosCompatibilityTest.java
@@ -220,7 +220,7 @@ class CosmosCompatibilityTest {
     }
 
     @Test
-    @DisplayName("transactional batch: Create/Read/Replace/Delete/Upsert")
+    @DisplayName("transactional batch: CRUD, Patch, and rollback")
     @SuppressWarnings("unchecked")
     void transactionalBatch() {
         String id = dbId();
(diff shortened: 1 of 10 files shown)


hectorvent:
Thank you, making batches genuinely atomic is the right fix and lifting it into `StorageBackend` so all four backends honour it is the right level to do it at.

Two things, both ours rather than yours. We merged #143 today, which added a `remove()` method across the same four backends, so this branch now conflicts on `StorageBackend`, `InMemoryStorage` and `PersistentStorage`.

More importantly, #167 appears to contain this entire change. The `StorageBackend` additions are identical between the two and both add the same three test files, just under different commit hashes, so they would not merge cleanly on top of each other.

Nothing wrong with the code here, it is a question of which PR carries it. If #167 supersedes this one, I would rather merge that and close this as included, so you do not have to rebase the same work twice.

thomhurst:
Superseded by #167, which includes this transactional-batch work together with the ETag-condition fixes. Closing this PR as included in #167.

PR #170, opened 2026-07-31
Fix Azure .NET Service Bus SDK compatibility

### LINKED ISSUE #159: [BUG] Service Bus: .NET SDK cannot connect in broker mode — MSSBCBS SASL unsupported
## Service

Service Bus

## Azure API Action

AMQP data plane from `Azure.Messaging.ServiceBus` 7.20.1 using the standard development-emulator connection string:

```text
Endpoint=sb://localhost:5673;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=devkey;UseDevelopmentEmulator=true;
```

## Expected behavior

A .NET `ServiceBusClient` can send and receive through a non-mocked Floci-AZ namespace, matching the official Azure Service Bus emulator contract.

## Actual behavior

Every .NET send fails during SASL negotiation, before CBS or entity access:

```text
Azure.Messaging.ServiceBus.ServiceBusException: None of the server sasl-mechanisms ([ANONYMOUS,PLAIN]) are supported by the client (MSSBCBS). (MessagingEntityNotFound).
```

Artemis advertises `ANONYMOUS` and `PLAIN`; the .NET Azure SDK selects Azure's `MSSBCBS` SASL mechanism when `UseDevelopmentEmulator=true`. The existing `ServiceBusCbsResponder` is never reached because negotiation fails first.

## Reproduction

Start 0.10.0 with a real broker, create a queue through the management API, then run:

```csharp

var connectionString =
    "Endpoint=sb://localhost:5673;SharedAccessKeyName=RootManageSharedAccessKey;" +
    "SharedAccessKey=devkey;UseDevelopmentEmulator=true;";

await using var client = new ServiceBusClient(connectionString);
await using var sender = client.CreateSender("orders");
await sender.SendMessageAsync(new ServiceBusMessage("hello"));
```

The same queue is created successfully and Artemis is listening on 5673. Failure occurs at `SendMessageAsync` with the exception above.

## Why this matters

This is a hard blocker for .NET/Aspire applications using the official `Azure.Messaging.ServiceBus` SDK. Even non-session queues cannot connect, independently of session support tracked in #123.

Suggested acceptance:

- `Azure.Messaging.ServiceBus` 7.20.x sends, receives, completes, abandons, and dead-letters in broker mode.
- Standard `UseDevelopmentEmulator=true` connection strings work without application-side SDK patches.
- A broker-mode .NET compatibility test runs in CI so the data plane cannot silently regress.

## Environment

- SDK: `Azure.Messaging.ServiceBus` 7.20.1
- Runtime: .NET 10, Docker Desktop on Windows
- Floci-AZ: Docker, `FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false`

thomhurst:
## Summary

- add the MSSBCBS SASL mechanism and correct CBS request/reply routing
- patch the pinned Artemis/Proton AMQP stack for Azure SDK link limits, lock tokens, delivery counts, and two-phase settlement
- provision Azure-compatible entity dead-letter subqueues
- add a .NET 10 compatibility suite using Azure.Messaging.ServiceBus 7.20.1 and run it against the real broker in CI

## Validation

- mvnw.cmd test (558 tests passed)
- real broker compatibility test: send, receive, complete, abandon/redeliver, dead-letter, and receive from SubQueue.DeadLetter

Closes #159

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -81,6 +81,12 @@ jobs:
               -e SERVICEBUS_AMQPS_PORT=5671
               -e SERVICEBUS_NAMESPACE=default
               -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
+          - test: sdk-test-dotnet
+            emulator_env: >-
+              -e FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false
+            extra_env: >-
+              -e SERVICEBUS_HOST=floci-az-servicebus-default
+              -e SERVICEBUS_AMQP_PORT=5672
           - test: sdk-test-node
             extra_env: >-
               -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
@@ -109,6 +115,7 @@ jobs:
             -e FLOCI_AZ_SERVICES_DOCKER_NETWORK=compat-net \
             -e FLOCI_AZ_TLS_ENABLED=true \
             -e FLOCI_AZ_HOSTNAME=floci-az \
+            ${{ matrix.emulator_env }} \
             floci-az:test-native
 
       - name: Wait for floci-az to be ready
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@ go.sum
 package-lock.json
 test-results.xml
 test-results/
+compatibility-tests/sdk-test-dotnet/bin/
+compatibility-tests/sdk-test-dotnet/obj/
 
 *.log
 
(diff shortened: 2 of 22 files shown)


hectorvent:
Thank you for this, and for adding a .NET compat suite alongside it. Wiring it into both the Makefile and the CI workflow in the same PR is exactly right.

I went and checked the mechanism against the .NET SDK rather than taking the issue's word for it, and your diagnosis is spot on. The SDK registers an anonymous SASL handler under the literal name `MSSBCBS` (`AmqpConnectionScope.cs`), so the broker has to offer that mechanism or SASL never completes, and only then does the real token exchange happen over `$cbs`. You implement both halves, the acceptor now advertises `MSSBCBS` and the responder answers PUT TOKEN with a 200.

**(blocking)** The thing holding this up is not the code. You have four Service Bus PRs open and this one is contained in #190 exactly, seven of its nine commits. My suggestion is to land #190 and close this as included, with #171 and #172 rebasing onto the result. Does that match your intent?

thomhurst:
Yes, that matches my intent. I’m happy for #190 to land first; once it merges, #170 can be closed as included and I’ll rebase #171 and #172 onto current `main`.

hectorvent:
included in #190

PR #171, opened 2026-07-31
feat(servicebus): add duplicate detection

### LINKED ISSUE #164: [FEAT] Service Bus: duplicate detection for queues and topics
## Service

Service Bus

## API Action / Feature

Queue/topic duplicate detection: `RequiresDuplicateDetection` plus `DuplicateDetectionHistoryTimeWindow`, enforced against `MessageId` on the AMQP data plane.

Azure documentation: https://learn.microsoft.com/azure/service-bus-messaging/duplicate-detection

## Why is this needed?

Applications using transactional outboxes can retry a send after an ambiguous acknowledgement; the broker-side `MessageId` window is the final guard against delivering the same event twice.

Floci-AZ 0.10.0 does not persist or echo these entity properties and the Service Bus docs list duplicate detection as out of scope. A local broker therefore accepts duplicate `MessageId` values that production Azure suppresses, preventing faithful validation of retry behavior.

Suggested acceptance:

- Parse, store, and echo `RequiresDuplicateDetection` and `DuplicateDetectionHistoryTimeWindow` on queues/topics.
- When enabled, accept the duplicate send but route only the first message within the configured window, matching Azure.
- Scope history per entity and expire entries after the configured window.
- Add broker-mode SDK compatibility coverage.

Related: session support #123 and per-entity delivery settings #125.

- [x] Yes
- [ ] No

thomhurst:
## Summary

- persist and echo queue/topic duplicate-detection settings with Azure-compatible defaults and validation
- add an Artemis broker plugin backed by its atomic duplicate-ID cache, with per-entity history-window eviction
- normalize Azure .NET emulator entity paths and resolve subscription queues without changing canonical topic topology
- cover queues and topics through `Azure.Messaging.ServiceBus` 7.20.1, including suppression and window expiry

## Validation

- `./mvnw test` — 562 passed
- `dotnet build compatibility-tests/sdk-test-dotnet/SdkTestDotnet.csproj --no-restore` — passed
- broker-mode .NET compatibility suite — 2 passed (settlement/DLQ plus queue/topic duplicate detection)

Depends on #170.

Closes #164

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
@@ -174,13 +174,11 @@ public sealed class ServiceBusCompatibilityTests
         await Assert.That(processedMessages[processorSessionB]).IsEqualTo("processor-b");
     }
 
-    [Fact]
-    public async Task DotnetSdkSuppressesDuplicateMessageIdsForQueuesAndTopics()
+    [Test]
+    [Timeout(50_000)]
+    public async Task DotnetSdkSuppressesDuplicateMessageIdsForQueuesAndTopics(
+        CancellationToken cancellationToken)
     {
-        using var timeout = CancellationTokenSource.CreateLinkedTokenSource(
-            TestContext.Current.CancellationToken);
-        timeout.CancelAfter(TimeSpan.FromSeconds(50));
-        CancellationToken cancellationToken = timeout.Token;
         const string serviceBusNamespace = "default";
         string queue = $"duplicate-{Guid.NewGuid():N}";
         await EnsureNamespace(serviceBusNamespace, cancellationToken);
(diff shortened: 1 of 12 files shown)


hectorvent:
Thank you, and the duplicate detection substance all checks out: keying on MessageId with first message winning, the window measured from enqueue with eviction after expiry, the PT20S to P7D validation bounds, and the 10 minute default all match Azure ([docs](https://learn.microsoft.com/en-us/azure/service-bus-messaging/duplicate-detection)). The end to end proof in the dotnet suite (duplicate suppressed, then accepted after window expiry) ran green against the real broker.

**(blocking)** We merged #190 as agreed, and that put this branch into conflict in six files. The good news is that the conflicts are almost entirely the shared base work that just landed, so a rebase onto main should shrink this PR down to the dedupe feature itself: the plugin, the XML validation, and folding your duplicate suppression test into the suite file that is now on main. While rebasing, would you also rename the PR title to a Conventional Commit (for example `feat(servicebus): add duplicate detection`)? With multiple commits the squash merge uses the PR title, same as on #190. One heads up: #172 touches the same files, so stacking the two rebases will save you from them breaking each other again.

That rebase plus the title is the only blocker from my side.

thomhurst:
Addressed: rebased onto `main` after #190, resolved the shared Service Bus changes while preserving session behavior, force-pushed the branch, and renamed the PR to `feat(servicebus): add duplicate detection`. Validation: `./mvnw test` (721 passed) and `dotnet build compatibility-tests/sdk-test-dotnet/SdkTestDotnet.csproj --no-restore` (0 warnings/errors). GitHub checks are green and the PR is mergeable.

hectorvent:
Thank you, this is exactly what was needed. The rebase brought the PR down to the dedupe feature itself, the title is right, and stacking #172 on this branch saved everyone a second round of conflicts.

I also verified the greens are real rather than taking the checkmarks on faith: on this head the Java Service Bus suite now runs un-mocked (16 tests executed against the broker) and the dotnet suite runs all 3, including your duplicate suppression test.

No blockers from my side, and nothing left on yours. This one goes in ahead of #172 per the stack.

hectorvent (approved):

PR #185, opened 2026-08-03
fix(core): omit Content-Type on bodyless error responses

### LINKED ISSUE #184: [BUG] HEAD error responses send Content-Type with no body, crashing the Azure SDK for C++
## Service

`blob` (also affects `queue` and `table`. The fix is central)

## Azure API Action

`Get Blob Properties` (HTTP HEAD) on a nonexistent blob. Any HEAD request that results in an error is affected, e.g. `Get Container Properties`.

## Expected behavior

A HEAD response carries no body (RFC 9110 §9.3.2: the server "MUST NOT send content in the response"), so it must not advertise a `Content-Type` describing content it cannot send.

Real Azure and Azurite both stay safe this way. Azurite gates both the content type and the body write on the request method, in `src/blob/generated/middleware/error.middleware.ts`:

```ts
if (err.contentType && req.getMethod() !== "HEAD") { res.setContentType(err.contentType); }
if (err.body && req.getMethod() !== "HEAD") { res.getBodyStream().write(err.body); }
```

Note the asymmetry: the status code and the header loop run regardless of method. `x-ms-error-code` is still expected, since it is the SDK's documented fallback for recovering the error code ([Status and error codes](https://learn.microsoft.com/en-us/rest/api/storageservices/status-and-error-codes2)).

Expected on the wire:

```
HTTP/1.1 404 Not Found
x-ms-error-code: BlobNotFound
(no Content-Type)
```

## Actual behavior

floci-az returns `Content-Type: application/xml` on a zero-byte response:

```
HTTP/1.1 404 Not Found
Content-Type: application/xml      <-- but the body is 0 bytes
x-ms-error-code: BlobNotFound
```

`AzureErrorResponse.toXmlResponse` sets the content type and the entity unconditionally; on HEAD, JAX-RS discards the entity but keeps the header.

This hard-crashes any caller using the Azure SDK for C++ (pod restart loop, not a catchable error). `StorageException::CreateFromResponse` (`sdk/storage/azure-storage-common/src/storage_exception.cpp`) chooses a parser by substring-matching the content-type header, then hands the buffer straight to `XmlReader` with no emptiness check:

```cpp
std::vector<uint8_t> bodyBuffer = std::move(response->GetBody());
auto xmlReader = _internal::XmlReader(
    reinterpret_cast<const char*>(bodyBuffer.data()), bodyBuffer.size());
```

On an empty buffer libxml2 returns rc=−1 and `XmlReader` throws `std::runtime_error("Failed to parse xml.")` — from inside the `RequestFailedException` constructor. The exception object is therefore never constructed, the `catch (RequestFailedException)` that would have mapped 404 not-found never fires, and the process calls `terminate()`.

The libxml2 message depends on the version: 2.11.8 reports `Extra content at the end of the document`, newer versions report `Document is empty`.

The same crash class applies to the JSON error path (`table`, `functions`): the SDK's JSON branch calls `json::parse(bodyBuffer)` equally unguarded, and additionally dereferences `["error"]["code"]` / `["error"]["message"]` without checking those keys exist.

## Reproduction

```bash
# start the emulator, then:
curl -X PUT "http://localhost:4577/devstoreaccount1/c1?restype=container"

# HEAD a blob that does not exist — note Content-Type on a bodyless 404:
curl -sD- -o/dev/null -I http://localhost:4577/devstoreaccount1/c1/missing
```

Against a C++ consumer, the equivalent of `BlobClient::GetProperties()` on a missing blob aborts the process instead of throwing `RequestFailedException`.

## Environment

- Floci-az version / image tag: `main` @ `eb0bf42`
- C++ SDK: `azure-sdk-for-cpp` (verified against upstream `main`), libxml2 2.11.8
- How you're running Floci-az: Docker (in-cluster) and `mvnw quarkus:dev`

mcarvalho-singlestore:
## Summary

Error responses to `HEAD` requests no longer advertise a `Content-Type`.

A HEAD response carries no body (RFC 9110 §9.3.2), but `AzureErrorResponse` set the content type unconditionally, so JAX-RS discarded the entity while keeping the header, leaving `content-type: application/xml` on a zero-byte response. That combination hard-crashes any caller using the Azure SDK for C++: `StorageException::CreateFromResponse` picks a parser by substring-matching the content type and hands the buffer to `XmlReader` without checking that it is non-empty, so the resulting `std::runtime_error` escapes the `RequestFailedException` constructor before any catch clause exists and calls `terminate()`. A `Get Blob Properties` on a missing blob crash-looped the consumer instead of surfacing a 404.

Fixed centrally in a `@ServerResponseFilter`, so it covers all 66 `toXmlResponse`/`toJsonResponse` call sites across blob, queue, table and functions. The JSON path needs it too: the SDK's JSON branch calls `json::parse(bodyBuffer)` equally unguarded.

Closes #184

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

**Incorrect behavior:** `HEAD /{account}/{container}/{missing-blob}` returned `Content-Type: application/xml` with a 0-byte body.

**Reference implementation:** Azurite gates *both* the content type and the body write on the request method, in `src/blob/generated/middleware/error.middleware.ts`:

```ts
if (err.contentType && req.getMethod() !== "HEAD") { res.setContentType(err.contentType); }
if (err.body && req.getMethod() !== "HEAD") { res.getBodyStream().write(err.body); }
```

Note the asymmetry: status code and the header loop run regardless of method. This PR mirrors that exactly, so the behavior is copied from Microsoft's own emulator rather than invented.

**Spec basis:**

- RFC 9110 §9.3.2: "The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, a server MAY omit header fields forwhich a value is determined only while generating the content." §15.3 adds that HEAD responses "never include content". A `SHOULD` with an explicit MAY-omit carve-out — omitting a content type that describes content we are forbidden to send is within spec.
- [Status and error codes](https://learn.microsoft.com/en-us/rest/api/storageservices/status-and-error-codes2) documents the `<Error><Code/><Message/></Error>` body and the `x-ms-error-code` header, but does **not** require `Content-Type` on error responses. `x-ms-error-code` is preserved - it is the SDK's documented fallback once body parsing is skipped, and is what lets the C++ SDK still map 404 not-found.
- [Get Blob Properties](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties) lists `Content-Type` among its **200** response headers, which is why the fix is scoped to  `status >= 400`. A successful HEAD keeps its content type.
**Scope.** Errors only, and only where a body cannot be sent. `GET` error responses are unchanged and still return the full `<Error>` document with `application/xml`.

## Checklist

## Testing

572 tests pass (Docker-backed suites excluded locally: `AksDockerTest`, `PostgresDockerTest`, `VmDockerTest` hang in `@AfterAll` container teardown on this machine, unrelated to this change; CI runs them).

The two HEAD-error tests were confirmed to **fail before the fix** with exactly the reported symptom:

```
Expected header "Content-Type" was not null, was "application/xml". Headers are:
Content-Type=application/xml
x-ms-error-code=BlobNotFound
```

Eight tests, covering both directions so the fix cannot be over- or under-applied:

| Test | Asserts |
|---|---|
| `BlobServiceTest.headMissingBlobOmitsContentType` | 404, no `Content-Type`, `x-ms-error-code: BlobNotFound` |
| `BlobServiceTest.headMissingContainerOmitsContentType` | 404, no `Content-Type`, `x-ms-error-code: ContainerNotFound` |
| `BlobServiceTest.getMissingBlobStillReturnsErrorBody` | GET 404 **keeps** xml + `<Code>BlobNotFound</Code>` |
| `BlobServiceTest.headExistingBlobKeepsContentType` | HEAD 200 **keeps** `Content-Type` |
| `QueueServiceTest.headMissingQueueOmitsContentType` | 404, no `Content-Type`, `x-ms-error-code: QueueNotFound` |
| `QueueServiceTest.getMissingQueueStillReturnsErrorBody` | GET 404 keeps xml + `<Code>QueueNotFound</Code>` |
| `TableServiceTest.headOnUnsupportedTableOperationOmitsContentType` | 501, no `Content-Type` (the JSON path) |
| `TableServiceTest.getMissingEntityStillReturnsErrorBody` | GET 404 keeps json + `ResourceNotFound` |

No C++ compatibility suite exists in this repo and adding one is out of scope, so coverage is via the unit/integration tests above plus the cited SDK source and spec pages (`CONTRIBUTING.md` asks authors to explain test choices).

### DIFF
--- a/src/main/java/io/floci/az/core/BodylessErrorContentTypeFilter.java
+++ b/src/main/java/io/floci/az/core/BodylessErrorContentTypeFilter.java
@@ -1,0 +1,37 @@
+package io.floci.az.core;
+
+import org.jboss.resteasy.reactive.server.ServerResponseFilter;
+
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerResponseContext;
+import jakarta.ws.rs.core.HttpHeaders;
+
+/**
+ * Strips {@code Content-Type} from error responses that cannot carry a body.
+ *
+ * <p>A HEAD response never includes content (RFC 9110 §9.3.2), so JAX-RS discards the entity while
+ * keeping the content type set by {@link AzureErrorResponse}. That leaves
+ * {@code content-type: application/xml} on a zero-byte response - a combination the Azure SDK for C++
+ * cannot survive: {@code StorageException::CreateFromResponse} parses the body whenever the content
+ * type contains {@code "xml"} (or {@code "json"}) without checking that the buffer is non-empty, and
+ * the resulting {@code std::runtime_error} escapes the {@code RequestFailedException} constructor
+ * before any catch clause exists, calling {@code terminate()}.
+ *
+ * <p>Azurite gates both the content type and the body write on the request method for this reason,
+ * leaving the status code and the remaining headers untouched. This filter mirrors that behavior.
+ * {@code x-ms-error-code} is deliberately preserved: it is the SDK's documented fallback for
+ * recovering the error code once body parsing is skipped.
+ *
+ * <p>Restricted to error statuses: {@code Get Blob Properties} documents {@code Content-Type} among
+ * its 200 response headers, so a successful HEAD must keep it.
+ */
+public class BodylessErrorContentTypeFilter {
+
+    @ServerResponseFilter
+    public void stripContentTypeOnBodylessErrors(ContainerRequestContext request,
+                                                 ContainerResponseContext response) {
+        if (response.getStatus() >= 400 && "HEAD".equalsIgnoreCase(request.getMethod())) {
+            response.getHeaders().remove(HttpHeaders.CONTENT_TYPE);
+        }
+    }
+}
--- a/src/test/java/io/floci/az/services/BlobServiceTest.java
+++ b/src/test/java/io/floci/az/services/BlobServiceTest.java
@@ -720,6 +720,59 @@ public class BlobServiceTest {
             .header("Content-MD5", equalTo("XrY7u+Ae7tCTyyK7j1rNww=="));
     }
 
+    // HEAD carries no body (RFC 9110 9.3.2), so an error response must not advertise a content type
+    // either: the Azure SDK for C++ parses the body whenever content-type contains "xml", and an empty
+    // buffer throws std::runtime_error out of the RequestFailedException constructor -> terminate().
+    // Azurite gates both the content type and the body on the method for the same reason.
+    @Test
+    void headMissingBlobOmitsContentType() {
+        given().put("/{account}/{container}?restype=container", ACCOUNT, CONTAINER);
+
+        given()
+            .when().head("/{account}/{container}/{blob}", ACCOUNT, CONTAINER, "no-such-blob.txt")
+            .then()
+            .statusCode(404)
+            .header("Content-Type", nullValue())
+            .header("x-ms-error-code", "BlobNotFound");
+    }
+
+    @Test
+    void headMissingContainerOmitsContentType() {
+        given()
+            .when().head("/{account}/{container}?restype=container", ACCOUNT, "no-such-container")
+            .then()
+            .statusCode(404)
+            .header("Content-Type", nullValue())
+            .header("x-ms-error-code", "ContainerNotFound");
+    }
+
+    // Counterpart guard: GET is allowed a body, so the <Error> document and its content type must stay.
+    @Test
+    void getMissingBlobStillReturnsErrorBody() {
+        given().put("/{account}/{container}?restype=container", ACCOUNT, CONTAINER);
+
+        given()
+            .when().get("/{account}/{container}/{blob}", ACCOUNT, CONTAINER, "no-such-blob.txt")
+            .then()
+            .statusCode(404)
+            .contentType(containsString("xml"))
+            .header("x-ms-error-code", "BlobNotFound")
+            .body(containsString("<Code>BlobNotFound</Code>"));
+    }
+
+    // Get Blob Properties documents Content-Type among its 200 response headers, so a successful HEAD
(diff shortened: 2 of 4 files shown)


hectorvent:
Thank you for your patience on this one, and apologies: the conflict is our doing, not yours. #134 merged about twenty minutes after I reviewed this, and it happened to add tests to the same three files.

The good news is that it is mechanical. Only `TableServiceTest.java` actually conflicts (the blob and queue files merge cleanly), and it is just two additive edits landing in the same two spots: the hamcrest import block, and the end of the class where you both appended tests. Keeping both sides is the whole resolution, there is no behaviour to reconcile.

I checked for a real clash and there is none. Your filter fires only on HEAD with a 4xx or 5xx, so it leaves #134's POST to 501 alone, and that one keeps its body and content type. The two changes complement each other nicely.

Would you rebase on main and keep both sets of tests? Nothing else needed from my side.

mcarvalho-singlestore:
Done @hectorvent.

hectorvent:
Thank you for the rebase, and this is ready from my side.

I ran the three service test classes locally this time rather than going on CI alone, 68 tests all passing, and I checked that your new test really does catch the bug by reverting just the header strip line. It fails with `Expected header "Content-Type" was not null, was "application/json"`, which is exactly the crash path the C++ SDK hits, so the test is doing real work.

Everything from the earlier read still holds: the filter is correctly narrow at `status >= 400` and HEAD only, so successful HEAD responses keep their content type, which matters because Get Blob Properties documents it among its 200 headers.

No blockers from my side.

hectorvent (approved):
Tests run locally 68/0 and the regression proven by reverting the header strip line. Filter is correctly scoped to HEAD plus 4xx/5xx. Approving.

PR #186, opened 2026-08-03
feat(compat): add a C++ SDK compatibility suite

mcarvalho-singlestore:
## Summary

Adds `sdk-test-cpp`, an Azure SDK for C++ compatibility suite. GoogleTest cases driving the real SDK clients, dependencies resolved by vcpkg in manifest mode, mirroring the Python/Java/Node suites.

The C++ SDK is uniquely sensitive to bodyless error responses. `StorageException::CreateFromResponse` picks a parser by substring-matching the `content-type` header and hands the body to `XmlReader` without checking that the buffer is non-empty. On a `HEAD` error response advertising `application/xml` with zero bytes, the resulting `std::runtime_error` is thrown from *inside* the `RequestFailedException` constructor — the exception object is never built, no catch clause can run, and the process calls `terminate()`.

Python, Java and Node all tolerate that response. This suite closes that blind spot.

> Depends on #185 (`fix/bodyless-error-content-type`). This branch is stacked on it so CI is green; `BlobErrors.GetPropertiesOnMissingBlobThrowsNotFound` fails without that fix.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Testing

**10 tests**, all passing against the emulator:

| Test | Covers |
|---|---|
| `BlobLifecycle.UploadDownloadDelete` | upload, download, delete round-trip |
| `BlobLifecycle.GetPropertiesOnExistingBlobSucceeds` | HEAD 200 keeps `Content-Type` |
| `BlobErrors.GetPropertiesOnMissingBlobThrowsNotFound` | 404  `BlobNotFound` |
| `BlobErrors.GetPropertiesOnMissingContainerThrowsNotFound` | 404  `ContainerNotFound` |
| `BlobErrors.DownloadMissingBlobParsesXmlErrorBody` | GET error keeps the parsable `<Error>` body |
| `BlobErrors.CreateExistingContainerThrowsAlreadyExists` | 409 `ContainerAlreadyExists` |
| `QueueLifecycle.EnqueueReceiveDelete` | enqueue, receive, delete |
| `QueueLifecycle.GetPropertiesOnExistingQueueSucceeds` | `ApproximateMessageCount` |
| `QueueErrors.GetPropertiesOnMissingQueueThrowsNotFound` | 404 `QueueNotFound` |
| `QueueErrors.ReceiveFromMissingQueueParsesXmlErrorBody` | queue XML error body |

**The regression test was verified in both directions.** Against an emulator image built with the
`BodylessErrorContentTypeFilter` removed, it fails with the reported symptom and the container exits
1:

```
[ RUN      ] BlobErrors.GetPropertiesOnMissingBlobThrowsNotFound
Entity: line 1: parser error : Document is empty
C++ exception with description "Failed to parse xml." thrown in the test body.
[  FAILED  ] BlobErrors.GetPropertiesOnMissingBlobThrowsNotFound
```

With the fix in place, 10/10 pass. So this is a test that demonstrably catches the bug rather than
one that merely passes.

## Build cost

Unlike the other suites there is no prebuilt package to install, so vcpkg compiles OpenSSL, libxml2, libcurl and the SDK from source.

| | Measured |
|---|---|
| Cold build | ~6 min |
| Warm rebuild (test edit) | 8s |
| Shipped image | 132 MB |
| Build stage | ~2.7 GB (discarded) |
| Test run | ~0.3s |

## CI

The matrix gains a per-suite timeout override:

```yaml
timeout-minutes: ${{ matrix.timeout_minutes || 20 }}
```

`sdk-test-cpp` gets 45; every other suite keeps the existing 20. The shared 20 min budget assumed
prebuilt dependencies, and the CI runners have fewer cores than my local machine, so a cold build
could possibly exceed it.

## Verification caveat

Everything above was measured on **x86_64**. CI runs on `ubuntu-24.04-arm`, and this machine has no arm64 emulation available, so I could not verify the arm64 build locally. Nothing is architecture-pinned: the CMake asset comes from `uname -m` and vcpkg auto-detects its triplet, but the arm64 build is genuinely first exercised by CI on this PR.

## Checklist

No emulator source is touched. This PR is additive test infrastructure, so the existing unit suite is unaffected.

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -68,7 +68,9 @@ jobs:
     name: native / ${{ matrix.test }}
     needs: build-native
     runs-on: ubuntu-24.04-arm
-    timeout-minutes: 20
+    # sdk-test-cpp compiles OpenSSL, libxml2, libcurl and the SDK from source via vcpkg,
+    # so it needs more headroom than the suites that install prebuilt packages.
+    timeout-minutes: ${{ matrix.timeout_minutes || 20 }}
     strategy:
       fail-fast: false
       matrix:
@@ -94,6 +96,8 @@ jobs:
           - test: sdk-test-node
             extra_env: >-
               -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
+          - test: sdk-test-cpp
+            timeout_minutes: 45
           - test: compat-terraform
           - test: compat-opentofu
           - test: compat-azcli
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -240,6 +240,7 @@ compatibility-tests/
                           managed identity
   sdk-test-node/        — Azure SDK for JS (jest): blob, queue, table, cosmos, appconfig,
                           keyvault, eventhub, managed identity
+  sdk-test-cpp/         — Azure SDK for C++ (GoogleTest via vcpkg): blob, queue. Built from source, so ~6 min cold
   compat-terraform/     — hashicorp/azurerm Terraform provider (BATS)
   compat-opentofu/      — OpenTofu with the azurerm provider (BATS)
   compat-azcli/         — real `az` CLI against a custom registered cloud (BATS)
@@ -272,6 +273,7 @@ Current per-suite env vars that must stay in sync:
 | `sdk-test-dotnet` | `-e SERVICEBUS_AMQP_PORT=5672` | ✓ |
 | `sdk-test-python` | `-e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577` | ✓ |
 | `sdk-test-node` | `-e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577` | ✓ |
+| `sdk-test-cpp` | none — only the shared `FLOCI_AZ_ENDPOINT` | ✓ (no `extra_env`) |
 
 The container name follows the pattern `floci-az-<service>-<namespace>` (e.g. `floci-az-servicebus-default`). If a new sidecar-based service is added, its container name and port must be added to both places.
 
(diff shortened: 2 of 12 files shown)


hectorvent:
Thank you for adding this, a C++ suite closes a real gap and the arm64 build in CI is a nice touch. The error codes the tests assert all line up with the published storage error enum, and the Makefile and workflow are updated together so nothing dangles between the local loop and CI.

**(blocking)** The one thing left is the PR title. This repo squashes with the "commit or PR title" setting, which uses the commit subject only when a PR has a single commit. This one has two, so the squash would land `Feat/sdk test cpp` on main, and semantic release cannot parse that, so the change would get no release note.

Worth saying explicitly: the Conventional Commits check is green because it lints your commits, and both of those are formatted correctly. It never looks at the title, so it cannot catch this.

Something like `feat(compat): add a C++ SDK compatibility suite` would do it. Also worth noting this stacks on #185, so that one wants to land first.

mcarvalho-singlestore:
Done @hectorvent

hectorvent:
Thank you for the title fix, that closes the release-notes concern completely. One new thing, and it is our doing: recent merges conflict with this branch in two files, the Makefile (the #190 merge, both add compat-suite targets) and TableServiceTest.java (the #182 merge). A quick rebase onto main should be the last step here; #185 is already merged so your base is in place.

mcarvalho-singlestore:
Done @hectorvent

hectorvent:
Thank you for the quick rebase. It came out clean: one conventional commit, mergeable, and the C++ job in CI actually ran all ten tests against the native image, which is the part I care most about for a new suite. The Makefile and workflow still move together (no per-suite env to mirror, and the timeout override lives only in the matrix where it belongs).

No blockers from my side, merging.

hectorvent:
Let's get it merge and see the time it takes. In past we had Rust in floci (aws) and it was taking too long.

hectorvent (approved):

PR #172, opened 2026-07-31
feat(servicebus): add entity message expiration

thomhurst:
## Summary

- Parse, validate, persist, and echo per-entity `DefaultMessageTimeToLive` settings.
- Apply the shorter of entity TTL and message TTL independently to queues and topic subscriptions.
- Move expired messages to entity DLQs when `DeadLetteringOnMessageExpiration` is enabled.
- Cancel scheduled expiry work when messages settle, move, expire, or their entity is deleted.

## Validation

- Focused Service Bus unit suite: 17 tests passed.
- Azure.Messaging.ServiceBus 7.20.1 against a real Artemis sidecar: entity TTL, shorter message TTL, receiver exclusion, and DLQ delivery passed.
- Existing PR checks passed before the follow-up; checks rerun after this push.

Depends on #170.

Closes #165

PR #190, opened 2026-08-08
feat(servicebus): support message sessions

### LINKED ISSUE #123: [FEAT] Service Bus: session support (RequiresSession entities + session-based receivers)
## Service

Service Bus

## API Action / Feature

Message sessions: `RequiresSession` on queues/subscriptions, `SessionId` grouping on the AMQP data plane, and session-based receive (`ServiceBusSessionProcessor` / `accept_session` across SDKs) with per-session lock + FIFO ordering.

The docs currently list sessions under "Out of scope (future work)" in `docs/services/service-bus.md` — this issue is a concrete real-world use case to help prioritize it.

## Azure Documentation

https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sessions

## Why is this needed?

We evaluated floci-az as a replacement for the official Service Bus emulator in a .NET Aspire app (multi-service event-driven architecture using Wolverine + `Azure.Messaging.ServiceBus`). Sessions were the single hard blocker: all ~14 of our consumer hosts use session-enabled subscriptions (`SessionId` = aggregate root id) so events for the same aggregate are processed in commit order. Every consumer runs a `ServiceBusSessionProcessor`, which cannot receive at all from a non-session entity.

Since the data plane is backed by ActiveMQ Artemis, Artemis message groups (`JMSXGroupID`) may be a viable mapping for the ordering/affinity half, though ASB session semantics (session lock, `SessionState`, accept-next-session) go a bit beyond that.

This would also differentiate floci-az meaningfully: the official Microsoft emulator supports sessions but is limited in other ways (heavy startup, config-file-frozen topology, connection limits), so session parity is what unlocks real-world event-driven apps switching over.

- [ ] Yes
- [x] No (happy to help test against a real-world Aspire + Wolverine workload)

thomhurst:
## Summary

- translate Azure AMQP session filters into Artemis message-group selectors with session ownership, FIFO ordering, accept-next-session, and lock metadata
- cover Java session receivers and .NET session receivers/processors against a real Artemis sidecar
- fix real-broker Java SDK compatibility for delivery counts, abandon, dead-letter/subscription path casing, and scheduled enqueue annotations
- run the Java Service Bus compatibility tests with mocked mode disabled in CI and the Makefile
- document supported session behavior and current limitations

## Validation

- `./mvnw test` — 601 passed
- Java `ServiceBusCompatibilityTest` — 12 passed against real Artemis
- .NET compatibility suite — 2 passed against real Artemis
- patched Artemis package build succeeded
- compatibility workflow YAML parsed successfully

Depends on #170.
Closes #123.

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -79,6 +79,7 @@ jobs:
           - test: sdk-test-java
             emulator_env: >-
               -e FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false
+              -e FLOCI_AZ_SERVICES_SERVICE_BUS_LOCK_DURATION_SECONDS=5
             extra_env: >-
               -e SERVICEBUS_HOST=floci-az-servicebus-default
               -e SERVICEBUS_AMQPS_PORT=5671
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - 'src/**'
       - 'pom.xml'
+      - 'Makefile'
       - 'docker/Dockerfile.native-package'
       - 'docker/entrypoint.sh'
       - 'compatibility-tests/**'
@@ -76,6 +77,8 @@ jobs:
             extra_env: >-
               -e AZURE_POD_IDENTITY_AUTHORITY_HOST=http://floci-az:4577
           - test: sdk-test-java
+            emulator_env: >-
+              -e FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false
             extra_env: >-
               -e SERVICEBUS_HOST=floci-az-servicebus-default
               -e SERVICEBUS_AMQPS_PORT=5671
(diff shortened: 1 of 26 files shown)


hectorvent:
Thank you, and I owe you a better read than my last one: I had assumed this mapped sessions onto Artemis message groups, and that is not what you have done.

Going through the .NET SDK, the session contract is stricter than message groups can satisfy. The client sets `com.microsoft:session-filter` even when no session id is given, then rejects the attach outright if the broker does not echo a concrete id back. Shipping patched protocol classes plus a real SASL provider looks like the only way to satisfy that, so the approach reads as necessary rather than heavy handed. Pinning the sidecar image to 2.44.0 to match the bundled patches is the detail that makes me comfortable with it.

**(blocking)** The title is not a Conventional Commit, and it is what gets used on squash merge, so it would land in the release notes as is. Something like `feat(servicebus): support message sessions` would do it.

I would like to land this first of your four, since it contains #170 outright and it is the change that finally makes the Service Bus suite run in CI. Does that work for you?

thomhurst:
Sounds good to me 😄

hectorvent:
Title looks great, thank you. That clears the one blocker, so no blockers from my side now. As agreed I will land this one first, then close #170 as included since all of its commits are in here, and #171 and #172 can rebase onto the result.

hectorvent (approved):

PR #195, opened 2026-08-13
fix(entra): make EntraStore participate in /_admin/reset

recchia:
## Summary
- `EntraStore` held mutable directory and Graph membership state (users, groups, memberships, authorization codes) but never implemented `Resettable`, so state written through the Entra/Graph API survived `POST /_admin/reset` instead of being wiped like every other stateful service.
- `clear()` now wipes the backing store and reseeds the zero-setup dev fixtures (tenant/app/user/group) so SDK clients keep working with no setup immediately after a reset.

Follow-up from #143, per @hectorvent's approval comment:
> `EntraStore` now holds mutable membership state but does not implement `Resettable`, so `/_admin/reset` leaves Graph writes behind. Worth a small follow up rather than holding this up.

## Test plan
- [x] `EntraStoreTest#clearWipesMembershipStateAndReseedsDevFixtures` — unit-level: a custom user/membership added after seeding is wiped by `clear()`, while the dev user/group/app/membership are reseeded.
- [x] `AdminResetTest#resetClearsGraphMembershipWrittenThroughEntraStore` — HTTP-level: a membership written via `POST /v1.0/groups/{id}/members/$ref` is gone after `POST /_admin/reset`, and the dev fixtures still resolve via `getMemberGroups` afterward.
- [x] Verified both new tests fail against the pre-fix code (confirms they exercise the reported bug).
- [x] `./mvnw -DskipITs -Dtest=EntraServiceTest,EntraStoreTest,GraphServiceHandlerTest,AdminResetTest test` — 29/29 green.

### DIFF
--- a/src/main/java/io/floci/az/services/entra/EntraStore.java
+++ b/src/main/java/io/floci/az/services/entra/EntraStore.java
@@ -3,6 +3,7 @@ package io.floci.az.services.entra;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import io.floci.az.config.EmulatorConfig;
+import io.floci.az.core.Resettable;
 import io.floci.az.core.StoredObject;
 import io.floci.az.core.storage.StorageBackend;
 import io.floci.az.core.storage.StorageFactory;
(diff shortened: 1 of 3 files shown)


hectorvent:
Thank you again, and this one is done from my side.

I ran the tests locally this time rather than going on CI alone: `AdminResetTest` and `EntraStoreTest` both pass, 9 tests with no failures. The run log is the useful part, it shows the admin reset clearing Entra alongside apim, appconfig, email and Graph in the same pass, then the dev fixtures being reseeded straight after, so the whole clear and reseed cycle is observably working rather than just asserted.

Everything from my earlier read still stands: the backend you clear is namespaced per service so the reset cannot reach other services' data, and `store` is the only mutable state on the class so nothing survives that should not.

No blockers from my side.

hectorvent (approved):
Backend isolation and state coverage verified from source, and the tests run clean locally (9/0). Approving.

PR #196, opened 2026-08-15
fix(functions): accept deployment packages larger than 20 MB

### LINKED ISSUE #192: [BUG] Functions deploy endpoint still hits 20 MB StreamReadConstraints limit (PR #41 only fixed Blob Storage)
## Service

az fn (Functions deploy endpoint)

## Azure API Action

`PUT devstoreaccount1-functions/admin/apps/{app}/functions/{fn}` (deploy function code via `zipBase64`)

## Expected behavior

Real Azure Functions zip deployment has no ~20 MB cap on package size - a Python function bundling several third-party SDKs (a few hundred MB of `.python_packages`) deploys and runs normally.

## Actual behavior

Any deploy whose `zipBase64` field exceeds 20,000,000 characters is rejected outright:

```
{"Code":"InvalidInput","Message":"Invalid request body: String value length (58721234) exceeds the maximum allowed (20000000, from `StreamReadConstraints.getMaxStringLength()`) (through reference chain: io.floci.az.services.functions.FunctionModels$DeployFunctionRequest[\"zipBase64\"])"}
```

This is the exact `StreamReadConstraints` issue already reported and fixed for Blob Storage in #37 / PR #41 ("fix(blob): support large blob uploads beyond 20 MB"). That fix does not appear to cover the Functions deploy endpoint - `FunctionModels$DeployFunctionRequest.zipBase64` is still parsed against Jackson's default `StreamReadConstraints.maxStringLength(20_000_000)`.

Encountered while deploying a Python v2-model function that bundles `databricks-sql-connector` (whose `client.py` does an unconditional `import pandas`, pulling in `pandas`+`numpy` at ~96 MB combined) - a real, legitimate dependency footprint that a small local team hit almost immediately.

## Reproduction

```bash
# Any app/function; the size of the base64 string is what triggers it, not the content.
python3 -c "
import base64, json
payload = {'handler': 'function_app.handler', 'timeoutSeconds': 60, 'zipBase64': base64.b64encode(b'0' * 16_000_000).decode()}
print(json.dumps(payload))
" > payload.json

curl -X PUT "http://localhost:4577/devstoreaccount1-functions/admin/apps/some-app/functions/some-fn" \
  -H "Content-Type: application/json" \
  --data @payload.json
# -> 400 InvalidInput, StreamReadConstraints.getMaxStringLength()
```

## Environment

- Floci-az version / image tag: 0.10.0 (`floci/floci-az:latest`)
- Java SDK version (if applicable): N/A (Python Functions worker)
- How you're running Floci-az (Docker / native / `mvn quarkus:dev`): Docker Compose

ankit1324:
## Summary

Closes #192

The `ObjectMapperCustomizer` added in #41 ("fix(blob): support large blob uploads beyond 20 MB") raises Jackson's string-length limit on the **CDI-managed** `ObjectMapper`. `FunctionsServiceHandler` constructs its own mapper in its constructor, so the customizer never reached it and it kept Jackson's default `maxStringLength` of 20,000,000 characters.

Any deploy whose `zipBase64` exceeded that was rejected outright:

```
{"Code":"InvalidInput","Message":"Invalid request body: String value length (58721234) exceeds the maximum allowed (20000000, from `StreamReadConstraints.getMaxStringLength()`) ... DeployFunctionRequest[\"zipBase64\"]"}
```

That is exactly the failure #192 reports, and it makes any function bundling a realistic dependency set (the reporter hit it with `pandas`+`numpy` at ~96 MB) undeployable.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Incorrect behavior: a zip deployment package over ~20 MB was rejected with `InvalidInput`. Real Azure Functions applies no such cap to the deployment package, so a function that deploys fine against Azure could not be deployed against the emulator at all.

The fix extracts the constraint setup from `JacksonConfig` into a reusable static helper and applies it to the Functions handler's mapper, so both paths share one definition of the limit (512 MB) rather than duplicating the constant.

Scope note: the decoded ZIP is written to disk by `FunctionCodeStore` and only its **path** is persisted in `FunctionDefinition`, so the base64 string is never re-serialised through the storage layer — this one mapper was the only remaining place the limit applied.

## Checklist

- [x] `./mvnw test` passes locally — **full suite: 633 tests, 0 failures, 0 errors**
- [x] New or updated integration test added — `FunctionsServiceHandlerMockedTest.deployAcceptsZipBeyondJacksonStringDefault` deploys a function with a 24,000,000-character `zipBase64` and asserts 201 + listability. Verified it is a genuine regression test: with the one-line fix reverted it fails with the exact `StreamReadConstraints.getMaxStringLength()` error from the issue.

### DIFF
--- a/src/main/java/io/floci/az/core/JacksonConfig.java
+++ b/src/main/java/io/floci/az/core/JacksonConfig.java
@@ -14,6 +14,17 @@ public class JacksonConfig implements ObjectMapperCustomizer {
 
     @Override
     public void customize(ObjectMapper mapper) {
+        applyLargePayloadConstraints(mapper);
+    }
+
+    /**
+     * Apply the large-payload stream-read constraints to a mapper.
+     *
+     * <p>The customizer above only reaches the CDI-managed {@link ObjectMapper}. Handlers that
+     * construct their own mapper keep Jackson's 20 MB default string limit unless they call this,
+     * which would reject payloads real Azure accepts.
+     */
+    public static void applyLargePayloadConstraints(ObjectMapper mapper) {
         mapper.getFactory().setStreamReadConstraints(
                 StreamReadConstraints.builder()
                         .maxStringLength(MAX_STRING_LENGTH)
--- a/src/main/java/io/floci/az/services/functions/FunctionsServiceHandler.java
+++ b/src/main/java/io/floci/az/services/functions/FunctionsServiceHandler.java
@@ -1,6 +1,7 @@
 package io.floci.az.services.functions;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.floci.az.core.JacksonConfig;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import io.floci.az.config.EmulatorConfig;
@@ -58,6 +59,10 @@ public class FunctionsServiceHandler implements AzureServiceHandler, Resettable 
         this.mapper    = new ObjectMapper()
                 .registerModule(new JavaTimeModule())
                 .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+        // This mapper is built here rather than injected, so the ObjectMapperCustomizer that
+        // raises Jackson's string-length limit does not reach it — without this a function ZIP
+        // over 20 MB is rejected, which real Azure Functions accepts.
+        JacksonConfig.applyLargePayloadConstraints(this.mapper);
     }
 
     @Override public String getServiceType() { return "functions"; }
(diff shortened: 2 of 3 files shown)


hectorvent:
Thank you, and the diagnosis here is exactly right: the customizer only reaches the CDI managed mapper, and this handler builds its own, so the raised limit never applied to it. Extracting the constraint into a shared static rather than forcing an injected mapper is the right call, since this one is configured differently.

I checked that your test really does cover the regression by reverting just your one line locally, and it fails with `String value length (20051112) exceeds the maximum allowed (20000000)` pointing straight at `zipBase64`. That is exactly the bug, so the test is doing real work rather than just exercising the path.

**(follow-up, separate PR)** Worth knowing the pattern is wider. Six classes read a request body through their own mapper, and `EmailHandler` is the one with the same exposure, since ACS attachments arrive as base64 in `contentInBase64` through an unconstrained static mapper. A shared mapper factory would close the whole class of bug.

No blockers from my side.

hectorvent (approved):
Regression proven by reverting the fix line locally: the test fails with the exact StreamReadConstraints error naming zipBase64. Tests 3/3. Approving.

PR #198, opened 2026-08-18
feat: add ADLS list paths support

snazy:
Summary:
- Add Blob-backed ADLS Path - List handling for DFS `listPaths` requests.
- Support recursive and non-recursive listing, directory filtering, deterministic continuation, and JSON path-list responses.
- Add SDK and HTTP coverage for `listPaths`, SAS list scope, and generic vended ADLS SAS property-map usage.

Type of change:
- [x] New feature (feat:)

Azure Compatibility:
- Implements the ADLS Path - List route used by Java Azure SDK DataLake clients: `GET /{filesystem}?resource=filesystem&recursive=...`
- SDK test source compile verified with repo-local Maven cache.

Checklist:
- [x] New or updated integration test added.
- [x] Commit messages follow Conventional Commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ flowchart LR
 
 | Service                 | Routing                      | Notable operations                                                                                                                                                                                                    |
 |-------------------------|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| **Blob Storage**        | `/{account}/`                | Create/delete containers, upload/download/delete blobs, list blobs; ADLS Gen2 DFS host alias, user delegation key vending, user delegation SAS enforcement                                                            |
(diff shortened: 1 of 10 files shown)


hectorvent:
Thank you, this is a clean one. I checked the shape against the Data Lake spec and the Java SDK and it holds up: continuation via the `x-ms-continuation` header, the 5000 page cap, and in particular serializing the entry etag as lowercase `etag`, which is what the SDK's generated Path model actually reads even though the swagger spells it `eTag`. Nice catch there, and the six DataLake SDK tests all ran for real in CI.

One follow up for a separate issue, not this PR: the DFS-path error bodies here are XML like the rest of our blob-backed ADLS surface, while real DFS answers the JSON error envelope. The SDK maps errors from the `x-ms-error-code` header so nothing breaks, but it is worth aligning the whole surface at some point.

No blockers from my side.

hectorvent (approved):

PR #201, opened 2026-08-20
fix(blob): preserve ADLS list directory identity

snazy:
Summary:
- Preserve boundary spaces when resolving the ADLS listPaths directory filter.
- Reject file paths supplied as listPaths directories with PathNotFound.

Type of change:
- [x] Bug fix (fix:)

Azure Compatibility:
- List Paths now preserves the requested directory identity and requires that the directory filter resolve to a directory.

Checklist:
- [ ] ./mvnw test passes locally (focused BlobServiceTest passes with JDK 25)

### DIFF
--- a/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
+++ b/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
@@ -603,7 +603,7 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
             return new AzureErrorResponse("FilesystemNotFound", "The specified filesystem does not exist.")
                     .toXmlResponse(Response.Status.NOT_FOUND.getStatusCode());
         }
-        if (directory != null && !directory.isBlank()
+        if (directory != null && !directory.isEmpty()
                 && !dataLakePathOperations.pathExists(request.accountName(), filesystem, directory)) {
             return new AzureErrorResponse("PathNotFound", "The specified path does not exist.")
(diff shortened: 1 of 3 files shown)


snazy:
@hectorvent I left the Greptile review changes for #198 left uncommitted on my disk. Sorry for the additional work.

hectorvent:
Thanks for carrying the #198 follow-ups over, this closes them cleanly.

Both changes read as the right call against the List Paths contract. The spec only says an error occurs if the directory does not exist ([DataLakeStorage.json#L2764](https://github.com/Azure/azure-rest-api-specs/blob/7efab4ca79331ad2ac48eb67fb944d030745918d/specification/storage/data-plane/Azure.Storage.Files.DataLake/stable/2025-05-05/DataLakeStorage.json#L2764)) and never trims the filter, so preserving the boundary spaces is the parity-correct move and the old `trim()` was ours to lose. Requiring the exact match to carry the directory resource type, while still recognising implicit directories through descendants, matches how a hierarchical namespace behaves, and `PathNotFound` is the established ADLS code for it ([error_codes.go#L73](https://github.com/Azure/azure-sdk-for-go/blob/5a4cb2123f1d1c66e3a7fb0e30e5f0a97c2c4c97/sdk/storage/azdatalake/datalakeerror/error_codes.go#L73)).

Ran `BlobServiceTest` on your head locally: 68 passed, including the two new cases.

No blockers from my side.

hectorvent (approved):

PR #202, opened 2026-08-20
fix(blob): honor user delegation SAS signing layouts by version

snazy:
Summary:
- Select the user-delegation SAS string-to-sign layout from sv, including legacy, encryption-scope, delegated-user, and request-constraint fields.
- Parse the version-dependent SAS fields and fail closed for malformed versions or user-bound tokens that cannot yet be identity-validated.

Type of change:
- [x] Bug fix (fix:)

Azure Compatibility:
- Matches the documented Blob and Data Lake user-delegation SAS layouts from 2018-11-09 through 2026-04-06.

Checklist:
- [ ] New or updated integration test added (unit parser coverage added)

### DIFF
--- a/src/main/java/io/floci/az/core/auth/StorageSasAuthorization.java
+++ b/src/main/java/io/floci/az/core/auth/StorageSasAuthorization.java
@@ -149,11 +149,13 @@ public class StorageSasAuthorization {
                 value(token.signedKeyStart()),
                 value(token.signedKeyExpiry()),
                 value(token.signedKeyService()),
-                value(token.signedKeyVersion()),
-                value(token.preauthorizedAgentObjectId()),
-                value(token.agentObjectId()),
-                value(token.correlationId())
+                value(token.signedKeyVersion())
         ));
+        if (layout.includesAgentFields()) {
+            fields.add(value(token.preauthorizedAgentObjectId()));
+            fields.add(value(token.agentObjectId()));
+            fields.add(value(token.correlationId()));
+        }
         if (layout.includesDelegatedUser()) {
             fields.add(value(token.delegatedUserTenantId()));
             fields.add(value(token.delegatedUserObjectId()));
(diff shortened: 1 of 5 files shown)


hectorvent:
Thanks for this, the version dispatch is exactly what the SAS code needed, and the repeated-query canonicalisation with a test pinning it is a nice touch.

**(blocking)** The LEGACY layout (sv 2018-11-09 through 2019-12-12) looks inverted against both references. The pre-2020-02-10 string-to-sign has no `saoid`/`suoid`/`scid` slots but does keep the snapshot slot: see Azurite's [generateBlobSASSignatureUDK20181109](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/authentication/IBlobSASSignatureValues.ts#L541) and the Java SDK's [`VERSION <= 2019-12-12` branch](https://github.com/Azure/azure-sdk-for-java/blob/e2eddb6460286bb37546d75b0aeac0eb78a24d89/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/util/BlobSasImplUtil.java#L398), which agree (20 fields). So LEGACY should skip the three agent fields and include the snapshot line. No existing test signs with a pre-2020 `sv`, so a small legacy-layout test alongside the fix would pin it. The other four layouts match Azurite exactly.

**(follow-up, separate PR)** `BlobServiceHandler.authorizeAppend` is added without a caller; fine to drop or wire up later.

Fixing the LEGACY layout unblocks the merge from my side.

hectorvent:
Thank you for turning that around so quickly, and the fix is in the right place. Gating the three agent fields behind the layout rather than the version keeps the table readable, LEGACY now signs the twenty field form with the snapshot slot intact, and the four newer layouts are untouched.

The `legacySas` helper is the part I appreciate most: nothing else in the suite signs with a pre 2020 `sv`, so without it this would have stayed unpinned.

No blockers from my side.

hectorvent (approved):

PR #207, opened 2026-08-22
feat(servicebus): honor per-entity MaxDeliveryCount and LockDuration

### LINKED ISSUE #125: [FEAT] Service Bus: per-entity MaxDeliveryCount (and lock duration) instead of one global setting
## Service

Service Bus

## API Action / Feature

Per-queue / per-subscription `MaxDeliveryCount` and `LockDuration`, honored from the entity-create payload on the management plane, rather than the current emulator-wide `FLOCI_AZ_SERVICES_SERVICE_BUS_MAX_DELIVERY_COUNT` / `FLOCI_AZ_SERVICES_SERVICE_BUS_LOCK_DURATION_SECONDS` env vars.

Related properties from the same create payload worth honoring at the same time: `DeadLetteringOnMessageExpiration`.

## Azure Documentation

https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues#maximum-delivery-count

## Why is this needed?

On real Azure (and the official emulator), `MaxDeliveryCount` is a property of each queue/subscription, declared at create time. Our app provisions its topology through the management API with `MaxDeliveryCount = 5` on every subscription plus `DeadLetteringOnMessageExpiration = true`, and our poison-message handling/alerting is tuned to that. An emulator-global default of 10 means locally-reproduced retry/dead-letter behavior diverges from production: a message that dead-letters after 5 attempts in prod gets 5 extra local attempts, which masks poison-message bugs in exactly the environment meant to catch them.

Since entity creation is already dynamic, persisting these per-entity values and applying them to the Artemis address/queue configuration (redelivery attempts -> DLA) seems like a natural fit.

- [ ] Yes
- [x] No

hectorvent:
## Summary

Queues and subscriptions now honor `MaxDeliveryCount` (1–2000) and `LockDuration` (PT1S–PT5M) from the entity-create payload instead of always using the namespace-wide `FLOCI_AZ_SERVICES_SERVICE_BUS_*` defaults. The resolved values are passed into the Artemis address settings so dead-lettering and session-lock expiry follow the entity's own configuration. Out-of-range values return a 400 ATOM error.

Closes #125

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Limits match the Service Bus management API (`MaxDeliveryCount` 1–2000, `LockDuration` max PT5M). Verified with the Java SDK (`ServiceBusCompatibilityTest`) and the .NET SDK (`ServiceBusCompatibilityTests.cs`): a queue created with `MaxDeliveryCount=2` dead-letters after the 2nd abandon. Non-session peek-lock expiry is still not enforced by the broker (documented).

## Checklist

https://claude.ai/code/session_012mQkJ7Daai9eaJzL2Z6XVs

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
@@ -307,6 +307,70 @@ public sealed class ServiceBusCompatibilityTests
         await deadLetterReceiver.CompleteMessageAsync(expired, cancellationToken);
     }
 
+    [Test]
+    [NotInParallel("servicebus-broker")]
+    [Timeout(90_000)]
+    public async Task MaxDeliveryCountDeadLettersWhenExceeded(CancellationToken cancellationToken)
+    {
+        const string serviceBusNamespace = "default";
+        string queue = $"dotnet-mdc-{Guid.NewGuid():N}";
+        await EnsureNamespace(serviceBusNamespace, cancellationToken);
+        await EnsureQueue(serviceBusNamespace, queue, maxDeliveryCount: 2, cancellationToken);
+
+        string connectionString =
+            $"Endpoint=sb://{ServiceBusHost}:{ServiceBusPort};" +
+            "SharedAccessKeyName=RootManageSharedAccessKey;" +
+            "SharedAccessKey=devkey;UseDevelopmentEmulator=true;";
+        await using var client = new ServiceBusClient(connectionString, new ServiceBusClientOptions
+        {
+            RetryOptions = { TryTimeout = TimeSpan.FromSeconds(30) }
+        });
+
+        await using ServiceBusSender sender = client.CreateSender(queue);
+        await sender.SendMessageAsync(new ServiceBusMessage("poison"), cancellationToken);
+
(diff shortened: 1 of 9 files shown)


hectorvent:
CI: `sdk-test-dotnet` failed on the new `MaxDeliveryCountDeadLettersWhenExceeded` — first receive reported `DeliveryCount == 2`.

Root cause is pre-existing, not from this change: real Service Bus sends the AMQP `delivery-count` header 0-based (prior failed attempts), and the .NET SDK ([`AmqpMessageConverter.cs`](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs) adds `+1` on receive; the Go SDK does the same in `message.go`) while Java/JS/Python expose the header verbatim. Stock Artemis emits a 1-based header, so .NET sees one more than Azure would.

Both new compat tests now assert what they are about — dead-lettering after the entity's `MaxDeliveryCount`, with the count increasing between attempts — instead of exact values. Aligning the broker's `delivery-count` with Azure's 0-based semantics is a separate Artemis-side fix.

Issue #59, opened 2026-06-13
Add Azure Container Instances (ACI) support


hectorvent:
## Title

Add Azure Container Instances (ACI) support

## Description

Implement Azure Container Instances support inside floci-az.

The purpose is to run real Docker containers per container group and report their state, IP, and logs through the ARM control plane — the Azure counterpart of AWS ECS / Fargate, and a building block for Container Apps.

ARM provider: `Microsoft.ContainerInstance/containerGroups`. Each container group maps to one or more real local Docker containers; the handler surfaces `instanceView`, restart policy, ports/IP, and `containerLogs`.

Suggested capabilities:

* Container group CRUD (ARM)
* Launch real Docker containers (image from `imageReference`, env, ports, resources)
* Power/state reporting via `instanceView` (Running / Terminated / Waiting)
* `GET .../containers/{name}/logs` (container logs)
* `restart` / `stop` / `start` group actions

floci-az notes: reuse the `aks` / `vm` container lifecycle — go through `ContainerBuilder` + `ContainerLifecycleManager`, never call `dockerClient` directly; Docker failures must degrade gracefully (non-fatal).

## Official Documentation

* https://learn.microsoft.com/en-us/azure/container-instances/
* https://learn.microsoft.com/en-us/rest/api/container-instances/

## Azure SDK for Java

* Management: `com.azure.resourcemanager:azure-resourcemanager-containerinstance` — https://learn.microsoft.com/en-us/java/api/overview/azure/resourcemanager-containerinstance-readme

## Azure REST API Specs

* https://github.com/Azure/azure-rest-api-specs/tree/main/specification/containerinstance

PR #208, opened 2026-08-22
feat(aci): add Azure Container Instances emulation (mocked ARM CRUD)

hectorvent:
## Summary

Adds `Microsoft.ContainerInstance/containerGroups` on the ARM lane: container group CRUD + list (resource-group and subscription scope), PATCH tag updates, `start`/`stop`/`restart` with the spec's LRO shapes, synchronous DELETE, container logs, and canned location catalogs. Introduces `core/arm/ResourceIndexContributor` so container groups show up in the resource-group `/resources` index that azurerm checks before deleting a group. Mocked mode only (no Docker); Docker-backed container groups follow in a separate PR.

Refs #59

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Shapes follow the `Microsoft.ContainerInstance` spec (LRO via Location-header polling; restart signals its LRO on 204, stop is synchronous). Read-backs are azurerm-provider-safe: `containers[*].ports` and `resources.requests` are always present (server defaults cpu 1.0 / memoryInGB 1.5 match the az CLI), enum casing is canonical, and secure environment values / registry passwords / secret-volume contents are never echoed. `exec`/`attach` return 501; `azureFile`/`gitRepo` volumes return 400.

## Checklist

- [x] New or updated integration test added (`AciHandlerTest`, 22 tests)

https://claude.ai/code/session_012mQkJ7Daai9eaJzL2Z6XVs


hectorvent:
Addressed the Greptile P1: `mocked=false` is not implemented in this PR, so the handler now logs a startup warning and the banner no longer advertises a port range for that mode; the docs say the flag is reserved for the container-backed follow-up. Re-squashed and force-pushed.

hectorvent:
Addressed the DELETE parity note from my review pass: the spec declares 200 with the resource body for an existing group and 204 for "resource does not exist", so the handler now returns 200 + the group's ARM body before removal and keeps 204 for the absent (idempotent) path. `deleteThenGet404` pins both codes. Re-squashed and force-pushed.

PR #210, opened 2026-08-26
fix(blob): default unusable blob content types to application/octet-stream

### LINKED ISSUE #211: [BUG] Get Blob Properties returns 500 when the stored Content-Type is empty or not a valid media type
## Service

Blob

## Azure API Action

`Get Blob Properties` (HEAD). `Get Blob` (GET) and `List Blobs` are affected by the same stored value.

## Expected behavior

`Get Blob Properties` returns `200` with the content type stored for the blob, or `application/octet-stream` when none was specified.

[Get Blob Properties](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties) documents the response header as "The content type that's specified for the blob. If no content type is specified, the default content type is `application/octet-stream`." An empty `x-ms-blob-content-type` means none was specified, so the default applies.

[Put Blob](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob) documents `Content-Type` as "Optional. The MIME content type of the blob. The default type is `application/octet-stream`" and defines no validation of the value.

## Actual behavior

`500 Internal Server Error` with `content-type: text/plain; charset=utf-8`, no `<Error>` body and no `x-ms-error-code`, so the client cannot classify the failure, and so SDKs surface it as an unknown server error. The blob is effectively unreadable: `Get Blob` fails the same way, and `List Blobs` reports `<Content-Type></Content-Type>`.

The content type is stored as sent, but JAX-RS parses it again while rendering the response, after the handler has already returned:

```
java.lang.IllegalArgumentException: Failed to parse media type
  at org.jboss.resteasy.reactive.common.headers.MediaTypeHeaderDelegate.internalParse(MediaTypeHeaderDelegate.java:77)
  at org.jboss.resteasy.reactive.common.headers.MediaTypeHeaderDelegate.parse(MediaTypeHeaderDelegate.java:53)
  at org.jboss.resteasy.reactive.common.headers.MediaTypeHeaderDelegate.fromString(MediaTypeHeaderDelegate.java:128)
  at org.jboss.resteasy.reactive.common.util.MediaTypeHelper.valueOf(MediaTypeHelper.java:27)
  at org.jboss.resteasy.reactive.common.headers.HeaderUtil.getMediaType(HeaderUtil.java:119)
  at org.jboss.resteasy.reactive.common.jaxrs.ResponseImpl.getMediaType(ResponseImpl.java:209)
  at org.jboss.resteasy.reactive.server.handlers.ResponseHandler.handle(ResponseHandler.java:78)
```

Because the throw happens in the response-rendering path, the usual `AzureErrorResponse` machinery never runs, which is why the 500 carries no Azure error payload.

## Reproduction

```bash
docker run -d --name floci-az-011 -p 14577:4577 floci/floci-az:0.11.0
B=http://127.0.0.1:14577/devstoreaccount1
curl -s -o /dev/null -X PUT "$B/c1?restype=container"

# empty x-ms-blob-content-type (curl's "header;" syntax sends it with no value)
curl -s -o /dev/null -X PUT -H 'x-ms-blob-type: BlockBlob' -H 'x-ms-blob-content-type;' --data 'hello' "$B/c1/marker"

curl -sI "$B/c1/marker"                       # HTTP/1.1 500 Internal Server Error
curl -s -o /dev/null -w '%{http_code}\n' "$B/c1/marker"          # 500
curl -s "$B/c1?restype=container&comp=list" | grep -o '<Content-Type></Content-Type>'
```

A non-empty but malformed value fails identically:

```bash
curl -s -o /dev/null -X PUT -H 'x-ms-blob-type: BlockBlob' \
  -H 'x-ms-blob-content-type: not-a-media-type' --data 'hello' "$B/c1/marker2"
curl -sI "$B/c1/marker2"                      # HTTP/1.1 500 Internal Server Error
```

The header being absent entirely is fine because it already defaults to `application/octet-stream`. Only a present-but-unusable value triggers this.

## Impact

Any client that sets the header with an empty value loses access to the blob's properties and content.

## Environment

- Floci-az version / image tag: `floci/floci-az:0.11.0` (reproduced above); also present on `main` at `4d39b93`.
- Java SDK version (if applicable): n/a, reproduced over raw HTTP, so it is SDK-independent
- How you're running Floci-az (Docker / native / `mvn quarkus:dev`): Docker (JVM and native images both)

mcarvalho-singlestore:
## Summary

`Get Blob Properties` (HEAD), `Get Blob` and `List Blobs` answer `HTTP 500` for any blob whose stored `Content-Type` is empty or not a valid media type. The blob becomes permanently unreadable, so every later request for it fails, including the metadata-only HEAD.

Currently, the content type is client-supplied and stored as is, but JAX-RS parses it again while rendering the response, so an unusable value throws after the handler has already returned:

```
java.lang.IllegalArgumentException: Failed to parse media type
  at org.jboss.resteasy.reactive.common.headers.MediaTypeHeaderDelegate.internalParse
  at org.jboss.resteasy.reactive.common.jaxrs.ResponseImpl.getMediaType(ResponseImpl.java:209)
  at org.jboss.resteasy.reactive.server.handlers.ResponseHandler.handle(ResponseHandler.java:78)
```

Because the throw happens in the response-rendering path, Quarkus emits a bare 500 with no Azure `<Error>` body and no `x-ms-error-code`.

This PR substitutes `application/octet-stream` (what Azure reports for a blob whose content type was never set) when the stored value is missing, empty, or unparseable.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Behavior on `PUT Blob` with `x-ms-blob-content-type`, then `HEAD` (Get Blob Properties):

| `x-ms-blob-content-type` sent | Azure | floci-az before | floci-az after |
| --- | --- | --- | --- |
| header absent | `200` `application/octet-stream` | `200` `application/octet-stream` | `200` `application/octet-stream` |
| present, empty | `200` `application/octet-stream` | **`500`** | `200` `application/octet-stream` |

- [Get Blob Properties](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties) documents the `Content-Type` response header as "The content type that's specified for the blob. If no content type is specified, the default content type is `application/octet-stream`." That default is what this fix falls back to.
- [Put Blob](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob) documents `Content-Type` as "Optional. The MIME content type of the blob. The default type is `application/octet-stream`", and specifies no validation of the value. An empty `x-ms-blob-content-type` therefore means "none specified" rather than "the empty media type", which is what makes `application/octet-stream` the correct answer for the empty case.

## Scope

Applied at the write boundary (`addBlobHttpProperties`, used by `Put Blob` and `Put Block List`) and again
where the stored value is emitted (`Get Blob`/`Get Blob Properties` header, and the `List Blobs`
`<Content-Type>` element).

## Checklist

## Tests

Five tests in `BlobServiceTest`, one per affected path; all five reproduce the 500 before the
fix:

| Test | Guards |
| --- | --- |
| `putBlobWithEmptyContentTypeServesOctetStream` | Get Blob Properties after `Put Blob` |
| `putBlobWithUnparseableContentTypeServesOctetStream` | the malformed, non-empty case |
| `getBlobWithEmptyContentTypeStillReturnsContent` | Get Blob still returns the payload |
| `listBlobsWithEmptyContentTypeReportsOctetStream` | the `<Content-Type>` element in List Blobs |
| `committedBlockBlobWithEmptyContentTypeServesOctetStream` | the Put Block List write path |

---

Resolves https://github.com/floci-io/floci-az/issues/211

### DIFF
--- a/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
+++ b/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
@@ -429,7 +429,7 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
                 .header("Last-Modified", RFC1123_DATE_TIME.format(so.lastModified()))
                 .header("ETag", so.etag())
                 .header("x-ms-blob-type", so.metadata().getOrDefault("BlobType", "BlockBlob"))
-                .header(HttpHeaders.CONTENT_TYPE, so.metadata().getOrDefault("Content-Type", "application/octet-stream"))
+                .header(HttpHeaders.CONTENT_TYPE, usableContentType(so.metadata().get("Content-Type")))
                 .header(HttpHeaders.CONTENT_LENGTH, contentLength)
                 .header("x-ms-blob-content-length", totalSize)
                 .header("Accept-Ranges", "bytes")
@@ -571,7 +571,7 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
                     RFC1123_DATE_TIME.format(so.lastModified()),
                     so.etag(),
                     (long) so.data().length,
-                    so.metadata().getOrDefault("Content-Type", "application/octet-stream"),
+                    usableContentType(so.metadata().get("Content-Type")),
                     so.metadata().getOrDefault("BlobType", "BlockBlob")
             ), includes(request.queryParams().get("include"), "metadata") ? userMetadata(so.metadata()) : null));
         });
@@ -787,13 +787,30 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
         if (contentType == null) {
             contentType = request.headers().getHeaderString(HttpHeaders.CONTENT_TYPE);
         }
-        metadata.put("Content-Type", contentType != null ? contentType : "application/octet-stream");
+        metadata.put("Content-Type", usableContentType(contentType));
         BLOB_HTTP_PROPERTY_HEADERS.forEach((requestHeader, responseHeader) -> {
             String value = request.headers().getHeaderString(requestHeader);
             if (value != null) {
                 metadata.put(responseHeader, value);
             }
         });
+    }
+
+    /**
+     * Returns a content type safe to send back, substituting {@code application/octet-stream} - the
+     * default {@code Get Blob Properties} documents for a blob with no content type specified - for one
+     * that is missing, empty, or not a valid media type.
+     */
+    private static String usableContentType(String contentType) {
+        if (contentType == null || contentType.isBlank()) {
+            return MediaType.APPLICATION_OCTET_STREAM;
+        }
+        try {
+            MediaType.valueOf(contentType);
+        } catch (IllegalArgumentException e) {
+            return MediaType.APPLICATION_OCTET_STREAM;
+        }
+        return contentType;
     }
 
     /**
--- a/src/test/java/io/floci/az/services/BlobServiceTest.java
+++ b/src/test/java/io/floci/az/services/BlobServiceTest.java
@@ -1289,6 +1289,104 @@ public class BlobServiceTest {
             .header("Content-Type", not(emptyOrNullString()));
     }
 
+    @Test
+    void putBlobWithEmptyContentTypeServesOctetStream() {
+        given().put("/{account}/{container}?restype=container", ACCOUNT, CONTAINER);
+
+        given()
+            .header("x-ms-blob-type", "BlockBlob")
+            .header("x-ms-blob-content-type", "")
+            .body(BLOB_CONTENT)
+            .when().put("/{account}/{container}/{blob}", ACCOUNT, CONTAINER, BLOB)
+            .then().statusCode(201);
+
+        given()
+            .when().head("/{account}/{container}/{blob}", ACCOUNT, CONTAINER, BLOB)
+            .then()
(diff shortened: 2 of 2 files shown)


hectorvent:
Thanks for chasing this down to the RESTEasy frame, the stack trace made the diagnosis obvious.

This matches the official emulator: Azurite resolves the content type as `x-ms-blob-content-type`, then the request `Content-Type`, then `application/octet-stream` ([BlockBlobHandler.ts#L49](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/handlers/BlockBlobHandler.ts#L49)), and an empty header is falsy there, so it lands on octet-stream exactly as this PR does. Normalising on read as well as on write is the part I like most: blobs already stored with an unusable value become readable again.

**(follow-up, separate PR)** For a present but unparseable value, Azurite keeps what the client sent, so real Azure probably echoes it back. A faithful echo needs a raw header write that skips the JAX-RS parse, so this is right for now; storing the raw value and normalising only on the way out would leave that door open.


No blockers from my side.

hectorvent (approved):

PR #220, opened 2026-08-26
fix(cosmos): enable .NET SDK queries

### LINKED ISSUE #216: [BUG] Cosmos DB: .NET SDK queries fail — empty queryEngineConfiguration ("queryengineConfiguration cannot be empty!")
## Service

Cosmos DB (SQL / NoSQL API)

## Azure API Action

Any query issued by the **.NET** `Microsoft.Azure.Cosmos` SDK in Gateway mode — `GetItemQueryIterator<T>(...)` for `SELECT`, `ORDER BY`, aggregates, cross-partition, `TOP`, etc.

## Expected behavior

Queries execute and return results, as they do for the Java/Python/Node SDKs.

## Actual behavior

Every .NET query throws **before any request is sent to the server**:
```
System.ArgumentException: queryengineConfiguration cannot be empty!
   at Microsoft.Azure.Cosmos.Query.Core.QueryPlan.QueryPartitionProvider..ctor(...)
```
Point operations are unaffected: `CreateItem`, `ReadItem`, `ReplaceItem`, `UpsertItem`, `DeleteItem`, ETag `IfMatch` (412), and item-TTL expiry all work. Only queries fail.

Probed against `0.11.0` with `Microsoft.Azure.Cosmos` 3.62.0, `ConnectionMode.Gateway`, `LimitToEndpoint = true`, HTTP/1.1: container create/replace (incl. composite indexes + `DefaultTimeToLive`) and all point ops passed; every `GetItemQueryIterator` call threw the above.

## Root cause

The .NET SDK parses queries **client-side** via the native ServiceInterop query engine, which it initialises from the `queryEngineConfiguration` field of the `DatabaseAccount` (gateway root) response. floci returns it empty:
```java
// CosmosHandler — DatabaseAccount response
body.put("queryEngineConfiguration", "{}");
```
The SDK's `QueryPartitionProvider` rejects an empty config with `ArgumentException: queryengineConfiguration cannot be empty!` and never proceeds to execute the query. The Java/Python/Node SDKs instead POST a server-side query-plan request (`x-ms-cosmos-is-query-plan-request: True`, which floci's `getQueryPlan` answers), so they are not affected — which is likely why this hasn't surfaced (the compatibility suite is Java/Python/Node).

## Reproduction

```csharp
using var client = new CosmosClient(
    "http://localhost:14577/devstoreaccount1-cosmos/", wellKnownEmulatorKey,
    new CosmosClientOptions { ConnectionMode = ConnectionMode.Gateway, LimitToEndpoint = true });
var container = client.GetContainer(db, coll);
var it = container.GetItemQueryIterator<MyDoc>(new QueryDefinition("SELECT * FROM c"));
await it.ReadNextAsync();   // throws ArgumentException: queryengineConfiguration cannot be empty!
```

## Impact

The .NET Cosmos SDK is unusable for any read that isn't a point read — i.e. all listing, filtering, ordering, and aggregation. Point-op-only apps work; anything query-driven does not.

## Suggested acceptance

- Return a valid, non-empty `queryEngineConfiguration` in the `DatabaseAccount` response (the shape the emulator/gateway normally advertises), so the .NET SDK's ServiceInterop path initialises and either (a) sends the parsed query to floci's engine, or (b) falls back to the server query-plan path floci already implements.
- Add a .NET SDK query compatibility test (`SELECT`, `ORDER BY`, `SELECT VALUE COUNT(1)`, cross-partition).

## Environment

- floci-az `0.11.0` (`/_floci/health` version `dev`).
- `Microsoft.Azure.Cosmos` 3.62.0, .NET 10, Gateway mode + HTTP/1.1.

thomhurst:
## Summary

- return an Azure-compatible queryEngineConfiguration for Cosmos account discovery
- support RID-based partition-key-range refreshes and ETag responses used by .NET
- handle .NET client query-planner envelopes, including multiple ORDER BY items
- add Microsoft.Azure.Cosmos compatibility and focused protocol coverage

Follow-up: #235 documents deliberate query-planner configuration differences.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature
- [ ] Breaking change
- [ ] Docs / chore

## Azure Compatibility

Supports the official .NET SDK query planning and gateway request shapes. Floci-AZ deliberately advertises sqlAllowLike=true and sqlDisableOptimizationFlags=0; #235 documents why.

## Checklist

- [x] Conventional Commits

Tests: focused 14 passed; Cosmos suite 116 passed; .NET build clean; live .NET query test passed.

Closes #216

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.ObjectModel;
 using Microsoft.Azure.Cosmos;
 using Newtonsoft.Json;
 
@@ -29,9 +30,15 @@ public sealed class CosmosCompatibilityTests
 
         try
         {
+            var properties = new ContainerProperties("items", "/pk");
+            properties.IndexingPolicy.CompositeIndexes.Add(
+                new Collection<CompositePath>
+                {
+                    new() { Path = "/rank", Order = CompositePathSortOrder.Ascending },
+                    new() { Path = "/id", Order = CompositePathSortOrder.Ascending }
+                });
             Container container = await database.CreateContainerAsync(
-                new ContainerProperties("items", "/pk"),
-                cancellationToken: cancellationToken);
+                properties, cancellationToken: cancellationToken);
             await container.CreateItemAsync(
                 new QueryItem("one", "a", 3),
                 new PartitionKey("a"),
@@ -57,6 +64,15 @@ public sealed class CosmosCompatibilityTests
             await Assert.That(ordered[0].Rank).IsEqualTo(1);
             await Assert.That(ordered[1].Rank).IsEqualTo(2);
             await Assert.That(ordered[2].Rank).IsEqualTo(3);
+
+            List<QueryItem> multiOrdered = await ReadAll(
+                container.GetItemQueryIterator<QueryItem>(
+                    "SELECT * FROM c ORDER BY c.rank, c.id"),
+                cancellationToken);
+            await Assert.That(multiOrdered.Count).IsEqualTo(3);
+            await Assert.That(multiOrdered[0].Id).IsEqualTo("two");
+            await Assert.That(multiOrdered[1].Id).IsEqualTo("three");
+            await Assert.That(multiOrdered[2].Id).IsEqualTo("one");
 
             List<int> counts = await ReadAll(
                 container.GetItemQueryIterator<int>("SELECT VALUE COUNT(1) FROM c"),
--- a/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
@@ -1,0 +1,97 @@
+using Microsoft.Azure.Cosmos;
+using Newtonsoft.Json;
+
+namespace FlociAz.Compatibility;
+
+[NotInParallel]
+public sealed class CosmosCompatibilityTests
+{
+    private const string CosmosKey =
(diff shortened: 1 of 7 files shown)


hectorvent:
Thanks, the grounded parts of this are strong. The account config string matches the real gateway payload key for key against the JS SDK fixtures, and the hierarchical RID change lines up with a captured Azure self link, where the database RID really is the collection RID's prefix. The RID addressed pkranges fallback is the right diagnosis for why .NET failed where Java did not.

**(blocking)** `splitTopLevel` tracks parentheses and quotes but not `[` or `{`, so the planner's multi key rewrite `[{"item": c.a}, {"item": c.b}] AS orderByItems` splits at the comma inside the array and mis projects rather than failing. The greedy `(.+)` in `DOTNET_ORDER_BY_ITEM_PATTERN` compounds it. Bracket depth plus iterating the items, with a two key ORDER BY case in the engine test and the .NET test, would close it.

**(follow-up, separate PR)** `sqlDisableOptimizationFlags` appears in no Azure source I could find, and `sqlAllowLike` is `true` here against `false` in every real payload; a line saying that divergence is deliberate would help.

**(follow-up, separate PR)** The body is missing the template's `Type of change`, `Azure Compatibility` and `Checklist` sections.

thomhurst:
Addressed multi-key planner envelopes in 4f8f12a: top-level splitting now tracks (), [], and {}, and ORDER BY items are decoded individually. Java focused tests and the .NET compatibility project pass. Follow-up docs: #235.

hectorvent:
Thank you, tracking bracket and brace depth alongside parentheses is the right fix, and it means the planner's multi key rewrite now survives the split instead of being cut at the comma inside the array. The added two key ORDER BY case pins it where it matters.

No blockers from my side. The two follow ups from last time still stand for whenever you are next in this area: `sqlDisableOptimizationFlags` has no Azure source I can find, and `sqlAllowLike` differs from every recorded payload, so a line recording both as deliberate would help the next reader. #235 already covers part of that.

hectorvent (approved):

PR #222, opened 2026-08-26
fix(cosmos): support .NET transactional batch

### LINKED ISSUE #217: [BUG] Cosmos DB: .NET SDK TransactionalBatch returns 400 — HybridRow binary batch body not parsed
## Service

Cosmos DB (SQL / NoSQL API)

## Azure API Action

`TransactionalBatch` from the **.NET** `Microsoft.Azure.Cosmos` SDK — `container.CreateTransactionalBatch(pk).PatchItem/ReplaceItem/DeleteItem(...).ExecuteAsync()`.

## Expected behavior

The batch executes with Azure's all-or-nothing semantics and per-operation results, as the JSON-array batch path already does for other clients (#161 / #167).

## Actual behavior

**Every** .NET transactional batch returns `400 BadRequest`, no operation applied. Probed against `0.11.0` with SDK 3.62.0:

| Batch | Result |
|---|---|
| `PatchItem(Set + Increment)` | 400, item unchanged |
| `PatchItem` with non-matching `FilterPredicate` (Azure: 412) | 400 |
| `ReplaceItem(existing)` + `DeleteItem(missing)` (Azure: 404, first op rolled back) | 400 / 400 (item unchanged — atomicity itself is fine) |
| `DeleteItem` with stale `IfMatchEtag` (Azure: 412) | 400 |

Non-batch point ops (including single-item ETag `IfMatch` → 412) work correctly, so this is specific to the batch endpoint.

## Root cause

The .NET SDK serializes a transactional batch body in the **HybridRow binary** format (`Microsoft.Azure.Cosmos.Serialization.HybridRow`, `ServerBatchRequest`/RecordIO), not as a JSON array. floci's `executeBatch` (`CosmosHandler`) reads the body as a JSON array:
```java
operations = MAPPER.readValue(raw, new TypeReference<List<Map<String,Object>>>() {});
```
so the binary HybridRow payload fails to parse → 400 for the whole batch. The Java/Python SDKs (and the curl repro in #161) send JSON, so the JSON path works for them; the .NET binary wire format is a distinct case not covered by #161/#167.

## Reproduction

```csharp
var batch = container.CreateTransactionalBatch(new PartitionKey("u1"))
    .PatchItem("id1", new[] { PatchOperation.Set("/kind", "z"), PatchOperation.Increment("/count", 10) });
using var resp = await batch.ExecuteAsync();
// resp.StatusCode == 400, item unchanged
```

## Impact

.NET apps that use `TransactionalBatch` (a common pattern for multi-item atomic writes and conditional patches) cannot use it against floci at all.

## Suggested acceptance

- Detect and parse the HybridRow/RecordIO binary batch body the .NET SDK sends (in addition to the existing JSON-array path), and return results in the format the .NET SDK expects.
- Add a **.NET** SDK `TransactionalBatch` compatibility test (create + patch + conditional patch + delete-missing atomicity), alongside the existing JSON-path coverage.

## Related

- #161 (transactional batch atomicity / Patch — JSON path) and #167 (ETag) addressed the JSON-array path; this issue is the distinct .NET **HybridRow binary** wire format.

## Environment

- floci-az `0.11.0` (`/_floci/health` version `dev`).
- `Microsoft.Azure.Cosmos` 3.62.0, .NET 10, Gateway mode + HTTP/1.1.

thomhurst:
## Summary

- decode and encode the RecordIO/HybridRow transactional-batch format used by Microsoft.Azure.Cosmos
- return 207 Multi-Status for failed HybridRow and JSON transactional batches
- preserve atomic rollback behavior
- evaluate batch patch FilterPredicate conditions
- add captured-wire and real .NET SDK coverage

Follow-ups: #230 handles standalone patch FilterPredicate; #233 pins Create, Read, and Upsert opcodes.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature
- [ ] Breaking change
- [ ] Docs / chore

## Azure Compatibility

Supports the official .NET SDK HybridRow batch wire format. Any failed transactional batch now returns 207 across SDK payload formats.

## Checklist

- [x] Conventional Commits

Tests: focused codec/batch passed; Cosmos suite 116 passed; .NET build and live transactional-batch test passed.

Closes #217

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
@@ -1,0 +1,107 @@
+using System.Net;
+using Microsoft.Azure.Cosmos;
+using Newtonsoft.Json.Linq;
+
+namespace FlociAz.Compatibility;
+
+[NotInParallel]
+public sealed class CosmosTransactionalBatchCompatibilityTests
+{
+    private static readonly string EmulatorEndpoint =
+        Environment.GetEnvironmentVariable("FLOCI_AZ_ENDPOINT") ?? "http://localhost:4577";
+
+    [Test]
+    [Timeout(60_000)]
+    public async Task DotnetSdkExecutesTransactionalBatches(CancellationToken cancellationToken)
+    {
+        string account = $"dotnetbatch{Guid.NewGuid():N}";
+        string databaseName = $"db-{Guid.NewGuid():N}";
+        string containerName = $"items-{Guid.NewGuid():N}";
+        using CosmosClient client = CreateClient(account);
+        Database database = await client.CreateDatabaseAsync(databaseName, cancellationToken: cancellationToken);
+        Container container = await database.CreateContainerAsync(
+            new ContainerProperties(containerName, "/tenant"),
+            cancellationToken: cancellationToken);
+
+        try
+        {
+            await container.CreateItemAsync(
+                new { id = "one", tenant = "u1", kind = "a", count = 1 },
+                new PartitionKey("u1"),
+                cancellationToken: cancellationToken);
+
+            using TransactionalBatchResponse patch = await container
+                .CreateTransactionalBatch(new PartitionKey("u1"))
+                .PatchItem("one",
+                [
+                    PatchOperation.Set("/kind", "z"),
+                    PatchOperation.Increment("/count", 10)
+                ])
+                .ExecuteAsync(cancellationToken);
+
+            await Assert.That(patch.StatusCode).IsEqualTo(HttpStatusCode.OK);
+            await Assert.That(patch.Count).IsEqualTo(1);
+            await Assert.That(patch[0].StatusCode).IsEqualTo(HttpStatusCode.OK);
+
+            ItemResponse<JObject> patched = await container.ReadItemAsync<JObject>(
+                "one", new PartitionKey("u1"), cancellationToken: cancellationToken);
+            await Assert.That(patched.Resource.Value<string>("kind")).IsEqualTo("z");
+            await Assert.That(patched.Resource.Value<int>("count")).IsEqualTo(11);
+
+            using TransactionalBatchResponse filteredPatch = await container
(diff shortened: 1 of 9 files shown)


hectorvent:
Thank you, this is unusually well evidenced. Three of the four codec tests decode genuine SDK emitted byte captures rather than the codec's own output, and the compat test proves rollback for real by re reading the item and asserting the pre batch values survived. Sniffing the body rather than branching on content type is what the SDK itself does in reverse.

**(blocking)** `int status = hybridRow && failed ? 207 : 200` gates the 207 on the .NET path, so a Java, Python or JS client gets 200 for a batch that failed and was fully rolled back. Both SDKs promote the operation status only on 207 ([BatchResponseParser.java#L124](https://github.com/Azure/azure-sdk-for-java/blob/e2eddb6460286bb37546d75b0aeac0eb78a24d89/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/batch/BatchResponseParser.java#L124)), so those callers see a success with nothing committed. It predates you, but this PR is the thing that establishes why 207 matters, and `CosmosTransactionalBatchTest` currently asserts the 200, so it would move with the fix.

**(follow-up, separate PR)** `condition` is honored only inside batches; a single `PatchItemAsync` with a non matching `FilterPredicate` still applies the patch instead of answering 412.

**(follow-up, separate PR)** The `Create`, `Read` and `Upsert` opcodes are neither pinned by a byte capture nor exercised by the test, unlike Patch, Replace and Delete.

thomhurst:
Addressed failed JSON batch status in a194127 and updated remaining assertions in 82c7f1f. Failed transactional batches now return 207 for both HybridRow and JSON. Focused transactional and item-condition suites pass. Follow-ups: #230 and #233.

hectorvent:
Thank you, that is the fix. Dropping the `hybridRow &&` gate means a Java, Python or JS client now gets the 207 too, so the operation status actually gets promoted and a rolled back batch stops looking like a success. Updating the three test files that pinned the old 200 was the part that mattered most, since those would have made the correct behaviour look like a regression later.

No blockers from my side.

hectorvent (approved):

PR #223, opened 2026-08-26
fix(servicebus): support session DLQ access

### LINKED ISSUE #213: [BUG] Service Bus: session-subscription dead-letter queue cannot be received from or peeked
## Service

Service Bus (AMQP data plane, broker mode)

## Azure API Action

Receiving from / peeking the **dead-letter sub-queue of a session-enabled subscription** (`SubQueue.DeadLetter`), after `DeadLetterMessageAsync` on the session receiver.

## Expected behavior

On real Azure and the official emulator, a dead-lettered message can be read back from the entity's `$DeadLetterQueue`. The dead-letter sub-queue of a session-enabled entity is received with an **ordinary (non-session) receiver** — this is the pattern the mature Wolverine Azure Service Bus transport uses (`ServiceBusReceiverOptions { SubQueue = SubQueue.DeadLetter }`), and it is how DLQ drain / redrive / poison-message inspection is built. `PeekMessagesAsync` on the DLQ should also work non-destructively.

## Actual behavior

Dead-lettering **into** the DLQ succeeds (`DeadLetterMessageAsync` returns), but the message cannot be read back by any method:

| Access method | Result |
|---|---|
| non-session receiver `CreateReceiver(topic, sub, {SubQueue = DeadLetter})` → `ReceiveMessagesAsync` | `InvalidOperationException: A session receiver is required for this entity` |
| session receiver on the DLQ entity path `AcceptSessionAsync("{topic}/Subscriptions/{sub}/$DeadLetterQueue", sessionId)` | `ServiceBusException (GeneralError)` |
| `PeekMessagesAsync` on the DLQ | `ServiceBusException (MessagingEntityNotFound): AMQ119002: target address {topic}/Subscriptions/{sub}/$DeadLetterQueue/$management does not exist` |

So a message can be dead-lettered but never received, peeked, or redriven — it is effectively lost.

## Root cause (from source, current `main`)

In `ServiceBusNamespaceManager.jolokiaCreateQueue`/`jolokiaCreateSubscription`, the DLQ is provisioned with the parent's session metadata (`applySessionMetadata(..., deadLetterQueue, requiresSession, ...)`), which makes the DLQ itself session-required — hence "A session receiver is required for this entity" for the ordinary receiver. On Azure the DLQ of a session entity does not impose session semantics on the DLQ receiver. Separately, the DLQ address has no `$management` node, so `peekMessages()` (the peek support added in #137) is unavailable on dead-letter sub-queues, and the session-receive path on the DLQ returns `GeneralError`.

## Reproduction

1. `floci/floci-az:0.11.0`, broker mode.
2. Create topic + subscription with `RequiresSession=true`.
3. Send a message with a `SessionId`, `AcceptSessionAsync`, receive it, `DeadLetterMessageAsync(msg, "reason", "desc")`.
4. Attempt to read the DLQ via any of the three methods above → each fails as tabulated.

## Impact

Blocks the entire dead-letter workflow for session-enabled subscriptions: DLQ depth inspection, draining, redrive, and poison-message assertions in tests. Combined with #(runtime counts are 0) it means the DLQ is write-only locally.

## Suggested acceptance

- An ordinary (non-session) `SubQueue.DeadLetter` receiver can receive and complete messages from a session-enabled subscription's DLQ (Wolverine's pattern), preserving `DeadLetterReason` / `DeadLetterErrorDescription` / original `SessionId` / `Subject` / application properties.
- `PeekMessagesAsync` works on the DLQ (give the DLQ address a `$management` node).
- Add SDK compatibility coverage for dead-letter round-trip on a session-enabled entity.

## Environment

- floci-az `0.11.0` (`/_floci/health` version `dev`); Artemis `2.44.0`.
- `Azure.Messaging.ServiceBus` 7.20.2, .NET 10.

thomhurst:
## Summary

- preserve session metadata while dead-lettering session messages
- acknowledge processed AMQP dead-letter commands with the SDK-compatible outcome
- persist dead-letter reason, description, and supported modified properties
- normalize unsupported AMQP Info values before writing Artemis properties
- add Java/.NET SDK and focused patch-JAR coverage

Follow-up: #231 enables session receivers directly on dead-letter queues.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature
- [ ] Breaking change
- [ ] Docs / chore

## Azure Compatibility

Matches Java and .NET SDK dead-letter command behavior and prevents unsupported AMQP types from aborting dead-letter processing.

## Checklist

- [x] Conventional Commits

Tests: focused Service Bus 62 passed; Java live suite 17 passed; .NET live suite 7 passed; property-conversion unit test passed.

Fixes #213

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
@@ -178,6 +178,59 @@ public sealed class ServiceBusCompatibilityTests
     }
 
     [Test]
+    [Timeout(90_000)]
+    public async Task SessionSubscriptionDeadLetterQueueSupportsPeekAndNonSessionReceive(
+        CancellationToken cancellationToken)
+    {
+        const string serviceBusNamespace = "default";
+        string topic = $"session-dlq-topic-{Guid.NewGuid():N}";
+        string subscription = $"session-dlq-sub-{Guid.NewGuid():N}";
+        string sessionId = $"session-{Guid.NewGuid():N}";
+        await EnsureNamespace(serviceBusNamespace, cancellationToken);
+        await EnsureTopic(serviceBusNamespace, topic, cancellationToken);
+        await EnsureSessionSubscription(
+            serviceBusNamespace, topic, subscription, cancellationToken);
+
+        string connectionString =
+            $"Endpoint=sb://{ServiceBusHost}:{ServiceBusPort};" +
+            "SharedAccessKeyName=RootManageSharedAccessKey;" +
+            "SharedAccessKey=devkey;UseDevelopmentEmulator=true;";
+
+        await using var client = new ServiceBusClient(connectionString, new ServiceBusClientOptions
+        {
+            RetryOptions = { TryTimeout = TimeSpan.FromSeconds(10) }
+        });
+        await using ServiceBusSender sender = client.CreateSender(topic);
+        var outgoing = new ServiceBusMessage("session dead letter")
+        {
+            MessageId = Guid.NewGuid().ToString("N"),
+            SessionId = sessionId,
+            Subject = "important"
+        };
+        outgoing.ApplicationProperties["tenant"] = "contoso";
+        await sender.SendMessageAsync(outgoing, cancellationToken);
+
+        await using (ServiceBusSessionReceiver sessionReceiver = await client.AcceptSessionAsync(
+            topic, subscription, sessionId, new ServiceBusSessionReceiverOptions(), cancellationToken))
+        {
+            ServiceBusReceivedMessage message = await Receive(sessionReceiver, cancellationToken);
+            await sessionReceiver.DeadLetterMessageAsync(
+                message, "validation-failed", "invalid payload", cancellationToken);
+        }
+
+        await using ServiceBusReceiver deadLetterReceiver = client.CreateReceiver(
+            topic, subscription, new ServiceBusReceiverOptions { SubQueue = SubQueue.DeadLetter });
+        ServiceBusReceivedMessage? peeked = await deadLetterReceiver.PeekMessageAsync(
+            cancellationToken: cancellationToken);
+        await Assert.That(peeked).IsNotNull();
+        await AssertDeadLetterProperties(peeked!, outgoing.MessageId, sessionId);
+
+        ServiceBusReceivedMessage received = await Receive(deadLetterReceiver, cancellationToken);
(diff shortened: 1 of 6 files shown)


hectorvent:
Thank you, and the evidence here is unusually strong: Microsoft's own live tests read a session enabled entity's DLQ with a plain non session receiver, in both the .NET and Java suites, so dropping the session gate is right. Echoing a bare `Rejected` is also what the .NET receiver needs, since it only throws when the echoed outcome carries an error, which is why this never bit Java.

**(blocking)** `ServiceBusDeadLetterSupport.apply` passes every `Info` value straight to `putObjectProperty`, but Artemis accepts only primitives, `String`, `SimpleString` and `byte[]` and throws otherwise, and that exception aborts the whole dead letter. Azure's own `DeadLetterMessagesSubscription` live test passes a `DateTime` in `propertiesToModify`, so it is reachable from documented usage. Converting or skipping unsupported types would cover it.

**(follow-up, separate PR)** A session receiver on the DLQ still fails, so it is worth a note that #213 does not cover that shape. The body is also missing the template's `Type of change`, `Azure Compatibility` and `Checklist` sections.

Housekeeping: #224 merged just now and this branch went conflicting against it, in `pom.xml`, `ServiceBusConfigGeneratorTest` and the .NET test file. Both sides are wanted in each spot, so a rebase keeping both should do it.

thomhurst:
Addressed unsupported AMQP Info values in 5395a6b by preserving Artemis-compatible primitives and normalizing Date/other unsupported types before property writes. Patch-JAR unit test passes. Follow-up session-DLQ receiver support: #231.

hectorvent:
Thank you, the conversion is the right shape: whitelist what Artemis genuinely accepts, turn a `Date` into an ISO instant rather than dropping it, and fall back to `toString()` for anything else, so a `DeadLetterMessageAsync` carrying a timestamp now completes instead of aborting the whole disposition. The new unit test loading the packaged patch jar and driving `configure` reflectively is a nice way to test the artifact that actually ships.

With its parent work merged, the delta here is now just the six files, and I read all of them. Nothing new to raise.

**(follow-up, separate PR)** `Character` is the one type Artemis accepts that the whitelist omits, so a char property becomes a string rather than staying a char. Harmless, and arguably fine, but worth a comment if it is deliberate.

**(follow-up, separate PR)** Transfer dead letter queues (`/$Transfer/$deadletterqueue`) are not handled, and the jar name in the test hardcodes the Artemis version, so a bump breaks it silently.

No blockers from my side.

hectorvent (approved):

PR #224, opened 2026-08-26
fix(servicebus): unpack AMQP batches

### LINKED ISSUE #212: [BUG] Service Bus: multi-message AMQP batch sends deliver only the first message
## Service

Service Bus (AMQP data plane, broker mode)

## Azure API Action

`ServiceBusSender.CreateMessageBatchAsync()` + `TryAddMessage(...)` + `SendMessagesAsync(batch)` from `Azure.Messaging.ServiceBus` (.NET), i.e. an AMQP transfer carrying multiple messages in the `0x80013700` (`amqp-sequence`/batched) message format.

## Expected behavior

A batch of N messages sent in one AMQP transfer is unpacked into N independent messages, each routed and delivered individually — matching the official emulator and real Azure. Batching is the **default** send path for several clients (e.g. Wolverine's Azure Service Bus transport sends everything through `CreateMessageBatchAsync`/`SendMessagesAsync`).

## Actual behavior

Only the **first** message of a multi-message batch is delivered; the remaining messages are silently lost. When the subscription also carries a rule/filter, the batch members' per-message application properties are not visible to the selector, so **none** of the batch is routed.

Probed against `floci/floci-az:0.11.0` (broker mode), `Azure.Messaging.ServiceBus` 7.20.2, Artemis `2.44.0`:

| Send | Received |
|---|---|
| `SendMessageAsync` (single) | 1 ✅ |
| batch of **1** (`CreateMessageBatch` + 1 × `TryAddMessage`) | 1 ✅ |
| batch of **3** | **1** ❌ (expected 3) |
| batch of 3, subscription has a `CorrelationFilter` on `Subject` | **0** ❌ (expected 3) |

The batch of 1 succeeding shows the batch *framing* is decoded; the batch of 3 delivering only the first shows the constituent messages 2..N are never extracted.

## Reproduction

1. Start `floci/floci-az:0.11.0` with `FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false`.
2. Create a topic + (session) subscription with no rule (accept-all) via the management API.
3. From .NET:
```csharp
await using var client = new ServiceBusClient(
    "Endpoint=sb://localhost:5673;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=devkey;UseDevelopmentEmulator=true;");
await using var sender = client.CreateSender(topic);

using var batch = await sender.CreateMessageBatchAsync();
for (var i = 0; i < 3; i++)
    batch.TryAddMessage(new ServiceBusMessage($"m{i}") { SessionId = "s1", MessageId = Guid.NewGuid().ToString() });
await sender.SendMessagesAsync(batch);

await using var r = await client.AcceptSessionAsync(topic, sub, "s1");
var got = await r.ReceiveMessagesAsync(10, TimeSpan.FromSeconds(4));
// got.Count == 1, expected 3
```

## Impact

This is a correctness-critical, silent data-loss bug for .NET producers. Any client that batches by default (Wolverine, and application code that calls `SendMessagesAsync` with more than one message) loses all but the first message per send, with no error surfaced to the sender. It does not reproduce with the single-message `SendMessageAsync` used in the docs' .NET example, which is likely why it hasn't surfaced — the Java/Python/Node compatibility suites also only send messages one at a time (or via non-SDK AMQP senders), so the batch path is not currently exercised.

## Suggested acceptance

- A multi-message AMQP batch (`0x80013700`) is unpacked into its constituent messages on the broker's AMQP ingress, each with its own application/system properties.
- Each unpacked message is routed independently, including through subscription rule selectors.
- Add a .NET (and ideally Java/Node) SDK compatibility test that sends a ≥3-message batch and asserts all are received.

## Environment

- floci-az `0.11.0` (Docker, `linux/amd64`); `/_floci/health` reports `version: dev`.
- Artemis sidecar `apache/activemq-artemis:2.44.0`.
- `Azure.Messaging.ServiceBus` 7.20.2, .NET 10.

thomhurst:
## Summary

- recognize Azure Service Bus batched-message-format transfers (0x80013700)
- decode each AMQP Data section into an independent broker message before routing
- preserve per-message bodies, subjects, application properties, and subscription filtering
- add a .NET SDK regression test that sends a three-message ServiceBusMessageBatch

Fixes #212

## Testing

- Maven focused Service Bus tests — 62 passed
- dotnet build in compatibility-tests/sdk-test-dotnet — 0 warnings, 0 errors
- live Docker + Artemis targeted .NET SDK regression — 1 passed
- live Docker + Artemis full .NET Service Bus compatibility suite — 7 passed

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/ServiceBusCompatibilityTests.cs
@@ -86,6 +86,54 @@ public sealed class ServiceBusCompatibilityTests
     }
 
     [Test]
+    [Timeout(90_000)]
+    public async Task DotnetSdkUnpacksBatchedMessagesBeforeSubscriptionRouting(
+        CancellationToken cancellationToken)
+    {
+        const string serviceBusNamespace = "default";
+        string topic = $"batch-topic-{Guid.NewGuid():N}";
+        string subscription = $"batch-sub-{Guid.NewGuid():N}";
+        await EnsureNamespace(serviceBusNamespace, cancellationToken);
+        await EnsureTopic(serviceBusNamespace, topic, cancellationToken);
+        await EnsureFilteredSubscription(
+            serviceBusNamespace, topic, subscription, "accepted", cancellationToken);
+
+        string connectionString =
+            $"Endpoint=sb://{ServiceBusHost}:{ServiceBusPort};" +
+            "SharedAccessKeyName=RootManageSharedAccessKey;" +
+            "SharedAccessKey=devkey;UseDevelopmentEmulator=true;";
+        await using var client = new ServiceBusClient(connectionString, new ServiceBusClientOptions
+        {
+            RetryOptions = { TryTimeout = TimeSpan.FromSeconds(10) }
+        });
+        await using ServiceBusSender sender = client.CreateSender(topic);
+        using ServiceBusMessageBatch batch = await sender.CreateMessageBatchAsync(cancellationToken);
+        for (var index = 0; index < 3; index++)
+        {
+            var message = new ServiceBusMessage($"message-{index}")
+            {
+                MessageId = Guid.NewGuid().ToString("N"),
+                Subject = "accepted"
+            };
+            message.ApplicationProperties["index"] = index;
+            await Assert.That(batch.TryAddMessage(message)).IsTrue();
+        }
+        await sender.SendMessagesAsync(batch, cancellationToken);
+
+        await using ServiceBusReceiver receiver = client.CreateReceiver(topic, subscription);
+        IReadOnlyList<ServiceBusReceivedMessage> received = await receiver.ReceiveMessagesAsync(
+            3, ReceiveTimeout, cancellationToken);
+        await Assert.That(received.Count).IsEqualTo(3);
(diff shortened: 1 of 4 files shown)


hectorvent:
This one is a pleasure to read. Every load bearing claim checks out against SDK source: the `0x80013700` batch format ([AmqpConstants.java#L23](https://github.com/Azure/azure-sdk-for-java/blob/e2eddb6460286bb37546d75b0aeac0eb78a24d89/sdk/core/azure-core-amqp/src/main/java/com/azure/core/amqp/implementation/AmqpConstants.java#L23)), one fully encoded message per `Data` section, and the `& 0xFFFFFFFFL` mask really is required, since Artemis widens the int message format and that constant sign extends. Deferring `AMQPStandardMessage` construction until after `decoder.setBuffer(null)` is subtle and correct, because the constructor takes the same thread local decoder. A one line comment there would stop someone simplifying it into a bug later.

**(follow-up, separate PR)** A Java suite batch test would be the more canonical anchor, since Java is the primary compat target and drives the identical wire format.

**(follow-up, separate PR)** The body is missing the template's `Type of change` and `Azure Compatibility` sections, and the latter is exactly where that `AmqpConstants` citation belongs.

No blockers from my side.

hectorvent (approved):

PR #233, opened 2026-08-27
test: cover Cosmos HybridRow batch opcodes

thomhurst:
## Summary

Add RecordIO/HybridRow fixtures that pin Create, Read, and Upsert transactional-batch opcode decoding.

Depends on #222. This PR temporarily includes its parent commits; the diff narrows after #222 merges.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Azure Compatibility

The .NET SDK uses opcodes 0, 2, and 20 for Create, Read, and Upsert. The decoder already maps them; this adds wire-format regression coverage.

## Checklist

Focused test: `./mvnw test -Dtest=CosmosHybridRowBatchCodecTest`

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
@@ -48,6 +48,26 @@ public sealed class CosmosTransactionalBatchCompatibilityTests
             await Assert.That(patched.Resource.Value<string>("kind")).IsEqualTo("z");
             await Assert.That(patched.Resource.Value<int>("count")).IsEqualTo(11);
 
+            using TransactionalBatchResponse opcodeBatch = await container
+                .CreateTransactionalBatch(new PartitionKey("u1"))
+                .CreateItem(new { id = "created", tenant = "u1", kind = "create" })
+                .ReadItem("one")
+                .UpsertItem(new { id = "upserted", tenant = "u1", kind = "upsert" })
+                .ExecuteAsync(cancellationToken);
+
+            await Assert.That(opcodeBatch.StatusCode).IsEqualTo(HttpStatusCode.OK);
+            await Assert.That(opcodeBatch.Count).IsEqualTo(3);
+            await Assert.That(opcodeBatch[0].StatusCode).IsEqualTo(HttpStatusCode.Created);
+            await Assert.That(opcodeBatch[1].StatusCode).IsEqualTo(HttpStatusCode.OK);
+            await Assert.That(opcodeBatch[2].StatusCode).IsEqualTo(HttpStatusCode.Created);
+
+            ItemResponse<JObject> created = await container.ReadItemAsync<JObject>(
+                "created", new PartitionKey("u1"), cancellationToken: cancellationToken);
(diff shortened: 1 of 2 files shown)


hectorvent:
Thank you for taking the follow-up seriously, but I do not think this pins the opcodes yet. The new fixture decodes cleanly, and the assertions are the right ones, so the problem is provenance rather than correctness: the byte string is reconstructible from the codec's own `OPERATION_TYPES` map and its own framing constants, and its payload data (`created`, `pk`, `value`) matches nothing else in the repo, where every other fixture in that file lines up with a specific call in the .NET compat test. If 20 were actually `Read` and 2 `Upsert`, this test would still pass, because the labels and the opcodes both come from that one map.

**(blocking)** The mapping can only be settled by real SDK traffic. Appending `.CreateItem(...)`, `.ReadItem(...)` and `.UpsertItem(...)` to the existing `CreateTransactionalBatch` chain in `CosmosTransactionalBatchCompatibilityTests.cs` would emit all three opcodes from the actual client, in the CI job that already runs green, and would fail loudly if any of the three integers were wrong.

Worth saying: this is a hard one to evidence, because `OperationType` is internal to a closed-source assembly and that SDK is not in our reference checkout. The gap is ours, and I am adding it to the manifest.

Housekeeping: #222 merged, and this branch went conflicting against it since it still carries the original commits. `git rebase --onto origin/main <#222's last commit here> <branch>` drops them and should narrow this to the one test.

hectorvent:
Thanks for the rebase, that resolved the conflict and the diff here is now just the one test, which makes it much easier to see.

The substance is unchanged though, and the ask from last time still stands. `CREATE_READ_UPSERT_CAPTURE` is the same constant, and I checked the compat suite again on current main: the `CreateItemAsync` and `ReadItemAsync` calls in `CosmosTransactionalBatchCompatibilityTests` are standalone container operations used for setup and verification, not operations inside a `TransactionalBatch`, so they travel the ordinary document path and never emit a HybridRow batch opcode. Nothing in the repo still emits 0, 2 or 20 from a real client.

**(blocking)** Adding `.CreateItem(...)`, `.ReadItem(...)` and `.UpsertItem(...)` to the existing `CreateTransactionalBatch` chain in that file is the whole fix: three lines, in a job that already runs green, and it would fail loudly if any of the three integers were wrong.

Ignore the rebase note on my earlier comment, that is done.

thomhurst:
Addressed in 7a730fd: the compatibility suite now sends real Azure Cosmos .NET SDK transactional-batch Create, Read, and Upsert operations and verifies results. The branch was rebuilt cleanly on current upstream main. Java codec tests: 5 passed; .NET build: succeeded.

hectorvent (approved):
Thank you, this is exactly the ask and it is cleared. The batch chain now sends real Create, Read and Upsert operations through the .NET SDK and checks the per operation status codes, so the opcode mapping is pinned by the client rather than by the codec's own table. If the emulator decoded a Create as a Read, the two follow up reads at the end would fail, which is the falsifiability the fixture on its own could not give.

The rebase helped too, the diff is now small enough to read at a glance.

One note for later, and it is ours rather than yours: the Java unit fixture is still a hand derived constant, so it only works as a fast local echo of the compat test. Worth remembering if that chain is ever trimmed.

No blockers from my side.

PR #234, opened 2026-08-27
test: cover Service Bus administration suffix routing

thomhurst:
## Summary

Cover .NET Service Bus administration entities whose names collide with registered account suffix routes.

Depends on #221. This PR temporarily includes its parent commits; the diff narrows after #221 merges.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Azure Compatibility

The parameterized regression test covers entity names ending in -table, -appconfig, -keyvault, and -functions when the .NET administration User-Agent and api-version marker are present.

## Checklist

Focused test: `./mvnw test -Dtest=ServiceBusServiceTest`

### DIFF
--- a/src/test/java/io/floci/az/services/ServiceBusServiceTest.java
+++ b/src/test/java/io/floci/az/services/ServiceBusServiceTest.java
@@ -2,6 +2,8 @@ package io.floci.az.services;
 
 import io.quarkus.test.junit.QuarkusTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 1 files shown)


thomhurst:
Fixed the failing Node Service Bus suite in cdd4453. AtomPub detection was taking precedence over the explicit /{account}-servicebus/ route, so the account prefix was not stripped and entity creation returned 404. Explicit Service Bus suffix routing now remains terminal; the focused ServiceBusServiceTest suite passes 20/20.

hectorvent:
Thank you, and this test is the real thing: it was rebuilt against main and all four parameterizations fail there, so it genuinely pins the routing change rather than passing either way.

**(blocking, and it belongs on #221 rather than here)** Chasing the gap this PR set out to close turned up a regression. A .NET administration client on the documented `/{account}-servicebus/{entity}` path answers **404** on this stack where main answers 201: `hasExplicitServiceBusAccountPrefix` is defined with `&& !isDotNetAdministrationClient`, so that client never gets the account prefix stripped and the whole first segment becomes the entity name. `docs/services/service-bus.md` documents that path as the way to create a queue. CI misses it because the .NET compat suite builds its admin client from `Endpoint=sb://{authority}`, so it only ever uses root paths.

**(blocking)** `dotNetAdministrationClientCanUseChildOfEntityEndingInServiceBusSuffix` currently pins the broken reading, asserting that `/orders-servicebus/Subscriptions/processor` addresses a topic literally named `orders-servicebus`. That one wants inverting along with the fix.

The three tests I mentioned last time are still untouched and still pass on main, so they guard the earlier heuristic rather than this one. Worth folding into the same pass.

thomhurst:
Rebased onto current upstream main, including the merged #221 routing fix. The parameterized suffix-routing coverage now runs against that corrected implementation; `ServiceBusServiceTest` passes 26/26.

hectorvent (approved):
Thank you, and this is cleared. The regression this PR surfaced was fixed on #221 before it merged, the test that pinned the broken reading is gone from main, and this branch is now byte for byte identical to main in the routing filter.

What is left here is the covering test, which is the right thing to land on its own. It is not redundant with the two cases already on main: those cover the `-servicebus` suffix itself, while these four cover the competing suffixes, and I traced each one through the current routing to confirm the 201 is real rather than incidental.

One thing still open from last time, and not yours to fix here: the two `AzureRoutingFilterTest` cases and the App Config guard still pass on main and still describe the older heuristic. Worth a separate pass some time.

No blockers from my side.

PR #235, opened 2026-08-27
docs: explain Cosmos query planner divergences

thomhurst:
## Summary

Document the deliberate sqlAllowLike and sqlDisableOptimizationFlags query-planner configuration differences.

Depends on #220. This PR temporarily includes its parent commits; the diff narrows after #220 merges.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Azure Compatibility

Explains why Floci-AZ advertises LIKE support and why Azure's internal optimization flags are fixed at zero rather than sourced.

## Checklist

- [ ] New or updated integration test added

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
@@ -1,3 +1,4 @@
+using System.Collections.ObjectModel;
 using Microsoft.Azure.Cosmos;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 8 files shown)


hectorvent:
Thank you, the `sqlAllowLike` bullet is exactly what I was hoping for: it names our value, names Azure's, and gives the reason, which is what makes a deliberate divergence readable years later.

**(blocking, one sentence)** The `sqlDisableOptimizationFlags` bullet explains the key by referring to "Azure's internal deployment-specific optimization flags", but that phrase does not correspond to anything I can find: the string appears in no Azure source, and the key is in none of the seven recorded gateway payloads, which otherwise match our map key for key. The accurate and more useful statement is the inverse, that this key appears in no observed Azure payload and we emit `0` so the advertised set stays a superset while nothing reads it.

**(follow-up, separate PR)** While you are in there, the divergence a reader will most want is that our map has twenty keys where Azure's has nineteen, the extra one being this key. Saying that outright is stronger than framing it as a value difference.

One accuracy note on the surrounding sentence: no SDK in any language actually reads `sqlAllowLike` outside test fixtures, so "must allow the SDK planner to submit those queries" is more than we can show. Something like "so the advertised capability set matches what the engine can execute" would be safe.

thomhurst:
Addressed in 8a19414: the docs now distinguish the observed 19-key Azure payload from Floci's 20-key map, state that sqlAllowLike matches engine capability, and avoid unsupported SDK-planner claims.

hectorvent:
Thank you, that is the accurate version. Saying that `sqlDisableOptimizationFlags` does not appear in the observed Azure payloads, and that we emit it so the advertised set stays a superset with nothing reading it, says what is actually true instead of implying Azure has flags of that name. It also folds in the point I had left as a follow up, that our map carries one key more than Azure's.

The `sqlAllowLike` bullet was already right, and the pair now reads as a proper record of two deliberate divergences, which is what a future reader auditing this against a real gateway payload will need.

No blockers from my side.

hectorvent (approved):

PR #236, opened 2026-08-28
feat(blob): support append blobs with add SAS permission

snazy:
## Summary
- Implement Append Block for existing Append Blobs and preserve append offsets and committed block counts.
- Map append-only user-delegation SAS permission a exclusively to Append Block.

## Type of change
- [x] New feature (feat:)

## Azure Compatibility
- Matches Blob user-delegation SAS Add permission semantics for Append Block.

## Checklist

### DIFF
--- a/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
+++ b/src/main/java/io/floci/az/services/blob/BlobServiceHandler.java
@@ -267,6 +267,8 @@ public class BlobServiceHandler implements AzureServiceHandler, Resettable {
                     response = putBlock(request, containerName, blobName);
                 } else if ("PUT".equalsIgnoreCase(method) && "blocklist".equals(comp)) {
                     response = putBlockList(request, containerName, blobName);
+                } else if ("PUT".equalsIgnoreCase(method) && "appendblock".equals(comp)) {
+                    response = appendBlock(request, containerName, blobName);
                 } else if (("GET".equalsIgnoreCase(method) || "HEAD".equalsIgnoreCase(method))
                         && "blocklist".equals(comp)) {
                     response = getBlockList(request, containerName, blobName);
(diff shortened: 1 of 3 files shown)


hectorvent:
Thanks, this is a clean implementation and the error shapes are right: 201 with the append offset and committed block count, 409 `InvalidBlobType` on a block blob and 404 on a missing one all match the official emulator down to the message string ([AppendBlobHandler.ts#L205](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/handlers/AppendBlobHandler.ts#L205)). Doing the whole read, validate and write inside the lease monitor is right too.

**(blocking)** `x-ms-blob-condition-appendpos` and `x-ms-blob-condition-maxsize` are accepted and ignored rather than answering 412 `AppendPositionConditionNotMet` / `MaxBlobSizeConditionNotMet` ([Azurite](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/persistence/LokiBlobMetadataStore.ts#L2467)). Append position is the concurrency guard for append blobs, so a retry loop that thinks it is protected will duplicate data and still see 201.

**(blocking)** `BlobCompDispatchTest` drops `appendblock` from its list rather than replacing it. That class exists to pin "an unimplemented comp must not be mistaken for a PutBlob", and its javadoc asks for the expectation to be replaced, so a case asserting `comp=appendblock` on a **block** blob returns 409 and leaves the content intact would keep the guard.

**(follow-up, separate PR)** Azurite accepts either `a` or `w` for Append Block; this accepts only `a`, so a `w`-only SAS gets a 403 it would not get from the emulator.

hectorvent:
Since the review, #272 (ADLS Gen2 DFS) landed and rewrote large parts of `BlobServiceHandler`, so this branch now conflicts there. That is on my merge order. When you pick this up again, a rebase onto main comes first, and the two points above (the `appendpos` and `maxsize` conditions, and replacing rather than dropping the `appendblock` dispatch expectation) still stand. Happy to help with the rebase if the DFS changes make it awkward.

snazy:
I've addressed the comments.
`w`-only SAS permission support is prepared in the (draft) commit-stack PR #121

hectorvent:
Thank you, and this is cleared. Both points are closed properly rather than narrowly. The `appendpos` and `maxsize` conditions now return 412 with the same codes and the same message text Azurite uses ([StorageErrorFactory.ts#L669](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/errors/StorageErrorFactory.ts#L669)), and the offset comparison is against the current blob length, which is the documented semantic.

The dispatch test is the part I appreciate most. Rather than dropping `appendblock` from the parameterised list, you gave it a case of its own asserting 409 `InvalidBlobType` and an intact blob, so the invariant is now stronger than the one it replaced.

No blockers from my side.

hectorvent (approved):

PR #250, opened 2026-08-31
feat(servicebus): start the default namespace on boot for a deterministic AMQP endpoint

### LINKED ISSUE #249: [FEAT] Service Bus: deterministic AMQP endpoint an orchestrator can model (sidecar port not discoverable)
## Service

Service Bus

## API Action / Feature

A deterministic, orchestrator-visible AMQP endpoint for the Service Bus data plane. Either:

1. serve/proxy Service Bus AMQP on a port declared on the **main floci-az container** (alongside `:4577`), or
2. let config pin the Artemis sidecar's published AMQP port (and surface it via the management API), so an orchestrator that started the main container can model the SB endpoint.

## Why is this needed?

As of 0.11.0, Service Bus AMQP terminates on the Artemis sidecar that floci-az spawns itself at runtime. An orchestrator (e.g. the CommunityToolkit .NET Aspire integration, `AddFlociAzure`) starts and models only the main container — it neither starts the sidecar nor knows its port, so:

- No `WithServiceBusReference` connection-string helper can exist. Blob works today because storage rides the main `:4577` HTTP endpoint (`AZURE_STORAGE_CONNECTION_STRING` is injected), and Cosmos is being added the same way (CommunityToolkit/Aspire#1545). Service Bus is the odd one out purely because its port isn't a modelable endpoint.
- Dynamic-port / isolated multi-instance scenarios (Aspire assigns random host ports; several AppHosts run concurrently) have no way to discover the SB endpoint at all.

This is the remaining structural blocker for first-class Aspire consumption of floci-az Service Bus; I'll file the matching connection-string helper on CommunityToolkit/Aspire referencing this issue.

- [x] Yes
- [ ] No

thomhurst:
Closes #249

## What

Adds `floci-az.services.service-bus.start-on-boot` (env `FLOCI_AZ_SERVICES_SERVICE_BUS_START_ON_BOOT`, default `false`). When set, the `default` namespace — and its Artemis sidecar, unless `mocked` — starts together with the emulator instead of on the first entity-management call, so the configured AMQP port is listening as soon as floci-az is up.

## Why

Investigating #249 showed the endpoint is already *deterministic*: `amqp-port`/`amqp-tls-port` pin the sidecar's host ports (5673/5674), and `GET /{account}-servicebus/namespaces` reports them. The remaining gap for orchestrators is the **lazy namespace start** — until something calls the management API, nothing listens on the AMQP port, which breaks health checks and clients that connect at startup, and leaves a hosting integration (e.g. the CommunityToolkit Aspire `AddFlociAzure`, CommunityToolkit/Aspire#1552) with no moment at which the endpoint it advertises is actually live.

With this flag, an orchestrator can treat the Service Bus endpoint like any other fixed-port service: set `FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false`, `..._START_ON_BOOT=true`, and optionally `..._AMQP_PORT=<port>`, then hand clients `Endpoint=sb://<host>:<port>;...;UseDevelopmentEmulator=true;`.

## Changes

- `EmulatorConfig.ServiceBusConfig`: new `startOnBoot()` (`@WithDefault("false")`).
- `ServiceBusContainerManager.onStart`: starts the default namespace after orphan reaping when the flag is set; `mocked` mode registers a mocked namespace instead. Failures log `ERROR` and never fail startup, per the sidecar rules.
- `ServiceBusNamespaceManager.DEFAULT_NAMESPACE` extracted as the shared constant (`ServiceBusHandler` now references it).
- `application.yml` + `docs/services/service-bus.md`: flag documented; new "Deterministic endpoint for orchestrators" section explains the pinned ports + `start-on-boot` + `GET /namespaces` discovery together.

## Tests

`ServiceBusContainerManagerTest` (+4): default stays lazy; broker-mode boot start with configured ports; mocked-mode boot registration; boot-start failure is non-fatal. 7/7 green locally.

Behavior with the flag unset is unchanged.

### DIFF
--- a/docs/services/service-bus.md
+++ b/docs/services/service-bus.md
@@ -95,6 +95,23 @@ Endpoint=sb://localhost:5673;SharedAccessKeyName=RootManageSharedAccessKey;Share
 `UseDevelopmentEmulator=true` tells the SDK to use plain AMQP (no TLS). The `SharedAccessKey` value is
 ignored — Artemis runs without authentication in dev mode.
 
+## Deterministic endpoint for orchestrators
+
+The AMQP data plane always binds the configured host ports (`amqp-port`, default `5673`;
+`amqp-tls-port`, default `5674`), so an orchestrator that starts floci-az knows the Service Bus
+endpoint up front — there is no dynamic port to discover. Two more pieces complete the story:
+
+- **`start-on-boot: true`** starts the `default` namespace (and its Artemis sidecar) together with
+  the emulator, instead of on the first entity-management call. Without it, nothing listens on the
+  AMQP port until a queue, topic, or namespace is created, which breaks health checks and clients
+  that connect at startup.
+- **`GET /{account}-servicebus/namespaces`** reports each running namespace's actual
+  `amqpPort`/`amqpsPort`, for tooling that wants to verify or discover the endpoint at runtime.
+
+This is what hosting integrations (e.g. .NET Aspire) should rely on: pass
+`FLOCI_AZ_SERVICES_SERVICE_BUS_AMQP_PORT` (with `..._MOCKED=false` and `..._START_ON_BOOT=true`),
+then hand clients `Endpoint=sb://<host>:<that port>;...;UseDevelopmentEmulator=true;`.
+
 ## Python SDK
 
 ```python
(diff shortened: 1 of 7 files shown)


hectorvent (approved):
Thank you, and the investigation is the valuable half here. Establishing that the endpoint was already deterministic, and that the real gap was the lazy start rather than the port allocation, is why this ends up as a small opt in flag instead of a rework.

The shape is right too: default off so nothing changes for anyone who does not ask for it, the boot start wrapped so a Docker less environment logs an error rather than failing startup, and a test that actually asserts that non fatal path rather than assuming it.

**(follow-up, separate PR)** The body uses What / Why / Changes / Tests rather than the repo template, so no Type of change box is ticked and there is no Azure Compatibility or Checklist section. Same on a few of your other open ones, so it is worth mentioning once rather than on each.

One thing for docs once #251 lands too: mounting a topology file starts the namespace at boot as a side effect, which is worth a sentence next to this flag.

No blockers from my side.

PR #253, opened 2026-08-31
fix: honor X-Forwarded-Proto in /metadata/endpoints so URLs are https behind a TLS proxy

### LINKED ISSUE #252: /metadata/endpoints echoes the inbound Host with an http:// scheme, blocking terraform azurerm custom environments behind a TLS proxy
**Environment:** `floci/floci-az@sha256:6566c73b8f4b6e3ef90322c8c84a47258bb1ca13c82ea1d40027f7af11247e30`
(`FLOCI_AZ_VERSION=0.11.0`), single exposed port `4577/tcp`, plain HTTP.

**What works — please don't change it:** the ARM control plane is genuinely
good. Unauthenticated PUT/GET round-trips succeed for resource group, VNet,
subnet, NSG, and storage account, and a fresh NSG comes back with the full
real-Azure set of six default security rules unprompted. This report is
about one scheme-handling detail, not the ARM implementation.

**The problem:** terraform's azurerm provider with `environment = "custom"`
fetches `GET /metadata/endpoints` over **HTTPS unconditionally**. Hitting
floci-az directly fails at that first hop:
`http: server gave HTTP response to HTTPS client`. Putting a
TLS-terminating reverse proxy in front (caddy, `reverse_proxy` with default
`Host` forwarding) clears that hop cleanly — but the metadata document then
comes back advertising the *inbound* `Host`, always prefixed `http://`:

- direct on `:14577` → `"resourceManager":"http://localhost:14577"`,
  `"loginEndpoint":"http://localhost:14577/"`
- through the TLS shim on `:14578` → `"resourceManager":"http://localhost:14578"`,
  `"loginEndpoint":"http://localhost:14578/"`

So the URLs *are* dynamic (the `Host` is echoed), but the scheme is
hard-coded `http://` regardless of how the request arrived. terraform then
takes `loginEndpoint` at face value and sends a **plaintext** token request
to the TLS-only listener, which rejects it:
`400 … Client sent an HTTP request to an HTTPS server`. The apply never
reaches a resource.

**Ask:** make the metadata document scheme-aware — honor
`X-Forwarded-Proto` when constructing self-referential URLs, and/or accept an
explicit public-base-URL setting (e.g. `FLOCI_AZ_PUBLIC_BASE_URL`). With an
https-aware metadata document, stock terraform azurerm should reach the ARM
plane behind a plain TLS shim, with **no change to floci-az's own listener**
and no TLS support needed inside floci-az itself.

MichaelDanCurtis:
## Summary

`GET /metadata/endpoints` built its self-referential URLs (`resourceManager`, `loginEndpoint`, `microsoftGraphResourceId`, `portal`, `gallery`, `audiences`) from `UriInfo.getBaseUri()`, whose scheme reflects the raw socket. Behind a TLS-terminating reverse proxy — the only way to give the azurerm provider the HTTPS metadata hop it demands, since floci-az serves plain HTTP on `4577` — the proxy→emulator hop is plaintext, so `UriInfo` sees `http` and every advertised URL came back `http://` even though the client reached the proxy over `https`.

This PR honors `X-Forwarded-Proto` to override **only the scheme** of the derived base URL (the `Host`-derived authority is unchanged), matching the `X-Forwarded-Proto` + connection-secure handling already present in the Cosmos/Postgres/Sql handlers. It takes the first value of a proxy-chain list and validates it to `http`/`https` before use.

Closes #252.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

**Incorrect behavior:** terraform's `azurerm` provider with `environment = "custom"` fetches `GET /metadata/endpoints` over HTTPS unconditionally, so reaching floci-az requires a TLS-terminating reverse proxy in front. That hop then cleared, but the returned metadata advertised `loginEndpoint: "http://<host>/"`; terraform took it at face value and sent a **plaintext** OAuth token request to the TLS-only listener, which rejected it with `400 … Client sent an HTTP request to an HTTPS server`. The apply never reached a resource.

With `X-Forwarded-Proto: https` set by the proxy, the document now advertises `https://` URLs and the token hop stays on TLS. Only the scheme is affected; direct (no-proxy) callers, and floci-az's own built-in TLS proxy, are unchanged (no `X-Forwarded-Proto` → prior behavior).

Note (out of scope for this PR, worth a follow-up): `RequestUrls.resolveBaseUrl(...)` — used by `EntraServiceHandler` for the OIDC `issuer`/`audiences` and by `EmailHandler` — has the same `X-Forwarded-Proto` gap, so the OAuth token issuer claims likely mis-scheme behind a proxy the same way. Kept out of this change to stay focused on #252; happy to file it separately.

## Checklist

- [x] Tests pass locally — the full `io.floci.az.core` package (routing, identity, TLS proxy, health, CORS, registry — **46 tests, 0 failures**) plus a full project compile, on JDK 25. Full disclosure: the complete multi-service `./mvnw test` spins up docker-based service emulators (Cosmos/ServiceBus/AKS/ACR/Redis) that I couldn't run in my sandbox, so I gated on the entire `core` package instead — the change is confined to `AzureIdentityController.metadataEndpoints` and is fully covered there. Happy to confirm a clean full-suite run if a maintainer wants it before merge.
- [x] New or updated integration test added — `AzureIdentityControllerTest` (4 cases: default scheme, `X-Forwarded-Proto: https` flips every URL, proxy-chain first value wins, garbage value ignored); existing `AzureRoutingFilterTest` stays green

### DIFF
--- a/src/main/java/io/floci/az/core/AzureIdentityController.java
+++ b/src/main/java/io/floci/az/core/AzureIdentityController.java
@@ -4,6 +4,7 @@ import io.floci.az.config.EmulatorConfig;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.*;
 import jakarta.ws.rs.core.Context;
+import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.core.UriInfo;
@@ -30,10 +31,28 @@ public class AzureIdentityController {
     @GET
     @Path("metadata/endpoints")
     @Produces(MediaType.APPLICATION_JSON)
-    public Response metadataEndpoints(@Context UriInfo uriInfo) {
+    public Response metadataEndpoints(@Context UriInfo uriInfo, @Context HttpHeaders headers) {
         // Derive base URL from the incoming request so the returned endpoints are
         // always reachable by the caller (host-network, Docker bridge, or TLS).
         String baseUrl = uriInfo.getBaseUri().toString().replaceAll("/+$", "");
+        // Honor X-Forwarded-Proto when a TLS-terminating reverse proxy fronts the
+        // emulator: the proxy->emulator hop is plaintext, so UriInfo sees "http",
+        // but the caller reached the proxy over "https" and the advertised URLs
+        // must reflect the CLIENT's scheme. Otherwise go-azure-sdk / terraform's
+        // azurerm custom-environment bootstrap takes loginEndpoint at face value
+        // and sends a plaintext token request to the TLS-only listener, which
+        // rejects it and the apply never reaches a resource. Only the scheme is
+        // overridden; the Host-derived authority from UriInfo is kept as-is.
+        // Mirrors the X-Forwarded-Proto handling in the Cosmos/Postgres/Sql handlers.
+        String forwardedProto = headers.getHeaderString("X-Forwarded-Proto");
+        if (forwardedProto != null && !forwardedProto.isBlank()) {
+            // A proxy chain may send a comma-separated list; the first value is
+            // the scheme the original client used.
+            String scheme = forwardedProto.split(",")[0].trim().toLowerCase();
+            if (scheme.equals("http") || scheme.equals("https")) {
+                baseUrl = baseUrl.replaceFirst("^https?://", scheme + "://");
+            }
+        }
         return Response.ok(Map.of(
             "name",                     "floci-az",
             // No trailing slash: in Azure Stack mode the go-azure-sdk concatenates this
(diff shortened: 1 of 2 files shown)


hectorvent (approved):
Thank you, this is a well aimed fix and the diagnosis in the description is the best part: tracing it from `loginEndpoint` coming back `http://` all the way to terraform sending a plaintext token request to a TLS only listener is what makes the change easy to agree with.

It matches the handling already in the Cosmos, Postgres and Sql handlers, and is actually stricter than all three, since none of them validates the value or handles a proxy chain. The test that feeds it a garbage scheme is the case most people leave out.

**(follow-up, separate PR)** This is now the fourth place decoding that header, and your own note about `RequestUrls.resolveBaseUrl` is right: I checked, it derives the scheme from the connection alone, so the OIDC `issuer` claim does mis-scheme behind a proxy. One shared helper would collapse all four and close that gap at the same time. Worth filing.

Thanks also for saying plainly which tests you could not run, rather than ticking the box.

No blockers from my side.

PR #270, opened 2026-09-02
fix: list every service's resources in the ARM resource index, at both scopes

### LINKED ISSUE #269: [BUG] resourceGroups/{rg}/resources omits Compute, ContainerService, ContainerRegistry, Cache and every DB provider
## Service

ARM (`Microsoft.Resources`) — the resource-group resource index.

## Azure API Action

`Resources - List By Resource Group` —
`GET /subscriptions/{sub}/resourceGroups/{rg}/resources?api-version=2021-04-01`

## Expected behavior

Real Azure returns **every** resource in the group, whatever provider owns it.

## Actual behavior

The listing returns only the providers `ArmHandler` knows about: its own inline Storage / Key Vault / Web state, plus Network, API Management, Managed Identity, plus whatever implements the CDI `ResourceIndexContributor` interface — and `AciHandler` is currently the only implementation.

So a resource group holding a VM comes back **without** the VM, even though the same VM answers a direct `GET` with `200` and appears in the type-scoped list under `/providers/Microsoft.Compute/virtualMachines`.

Providers with a management plane in floci-az that never reach the index:

| Provider | Handler |
|---|---|
| `Microsoft.Compute/virtualMachines` | `VmHandler` |
| `Microsoft.ContainerService/managedClusters` | `AksHandler` |
| `Microsoft.ContainerRegistry/registries` | `AcrHandler` |
| `Microsoft.Cache/Redis` | `RedisHandler` |
| `Microsoft.DBforPostgreSQL/flexibleServers` | `PostgresHandler` |
| `Microsoft.DBforMySQL/flexibleServers` | `MySqlHandler` |
| `Microsoft.DBforMariaDB/servers` | `MariaDbHandler` |
| `Microsoft.Sql/servers` | `SqlHandler` |

## Reproduction

```bash
docker run -d --rm --name floci-az-repro -p 4577:4577 \
  floci/floci-az@sha256:6566c73b8f4b6e3ef90322c8c84a47258bb1ca13c82ea1d40027f7af11247e30

B=http://localhost:4577
SUB=00000000-0000-0000-0000-000000000001
RG=repro-rg

curl -s -X PUT "$B/subscriptions/$SUB/resourceGroups/$RG?api-version=2021-04-01" \
  -H 'Content-Type: application/json' -d '{"location":"eastus"}'

curl -s -X PUT "$B/subscriptions/$SUB/resourceGroups/$RG/providers/Microsoft.Compute/virtualMachines/repro-vm?api-version=2024-11-01" \
  -H 'Content-Type: application/json' -d '{
    "location":"eastus",
    "properties":{
      "hardwareProfile":{"vmSize":"Standard_D2s_v3"},
      "storageProfile":{"imageReference":{"publisher":"Canonical","offer":"0001-com-ubuntu-server-jammy","sku":"22_04-lts","version":"latest"},
                        "osDisk":{"createOption":"FromImage","name":"repro-osdisk"}},
      "osProfile":{"adminUsername":"azureuser","computerName":"reprovm"}}}'

curl -s -X PUT "$B/subscriptions/$SUB/resourceGroups/$RG/providers/Microsoft.Storage/storageAccounts/reprosa?api-version=2023-01-01" \
  -H 'Content-Type: application/json' -d '{"location":"eastus","sku":{"name":"Standard_LRS"},"kind":"StorageV2"}'

# The VM exists and is listed at the type-scoped level:
curl -s -o /dev/null -w 'direct GET  -> HTTP %{http_code}\n' \
  "$B/subscriptions/$SUB/resourceGroups/$RG/providers/Microsoft.Compute/virtualMachines/repro-vm?api-version=2024-11-01"
curl -s "$B/subscriptions/$SUB/resourceGroups/$RG/providers/Microsoft.Compute/virtualMachines?api-version=2024-11-01" \
  | python3 -c 'import sys,json; print("type list  ->", [r["name"] for r in json.load(sys.stdin)["value"]])'

# ...but the resource-group aggregation does not contain it:
curl -s "$B/subscriptions/$SUB/resourceGroups/$RG/resources?api-version=2021-04-01" \
  | python3 -c 'import sys,json; print("rg index   ->", [(r["name"], r["type"]) for r in json.load(sys.stdin)["value"]])'
```

Output on the pinned release above:

```
direct GET  -> HTTP 200
type list  -> ['repro-vm']
rg index   -> [('reprosa', 'Microsoft.Storage/storageAccounts')]
```

The storage account is there. The VM is not.

## Impact

- `ResourceIndexContributor`'s own javadoc records why this listing matters: *"The azurerm provider calls that listing before deleting a resource group to verify it is empty; a service that skips registration lets `terraform destroy` remove a group whose resources still exist."* That is exactly the state VMs, AKS clusters, registries, Redis caches and every DB server are in today.
- Anything that enumerates a group generically rather than per-type — `az resource list -g`, the Resource Management SDKs, drift/verification tooling that diffs deployed state against expected — sees an incomplete group and reports real resources as missing.

## Root cause

`ArmHandler.handleResourceGroupBranch` builds the index from a hand-maintained set of subsystems plus `Instance<ResourceIndexContributor>`. The interface exists precisely so a service can opt in with one method, but only ACI ever did — every service added since (and every service that predates the interface) is silently absent.

## Related, not covered here

`GET /subscriptions/{sub}/resources` (subscription-level) returns only Key Vaults and API Management services, so on an estate with neither it returns an empty list even when the subscription holds resources. Same family of gap, different endpoint; happy to file separately.

## Environment

- floci-az version / image tag: `floci/floci-az@sha256:6566c73b8f4b6e3ef90322c8c84a47258bb1ca13c82ea1d40027f7af11247e30`
- Also reproduced against `main` at `a9e4799` with a `@QuarkusTest` + RestAssured case
- How you're running floci-az: Docker, and `./mvnw test`

MichaelDanCurtis:
## Summary

Both of ARM's generic resource listings were incomplete, for the same reason: nothing assembled them from a single source of truth about what the estate holds.

**`GET .../resourceGroups/{rg}/resources`** built its answer from a hand-maintained set — `ArmHandler`'s own Storage / Key Vault / Web state, plus Network, API Management and Managed Identity — plus the CDI `ResourceIndexContributor` lane, where `AciHandler` was the only implementation. Every other service with a management plane was missing from its own resource group.

**`GET /subscriptions/{sub}/resources`** aggregated only Key Vaults and API Management services, so a subscription holding neither answered `{"value":[]}` while its groups held VMs, virtual networks, storage accounts and database servers.

This PR registers the eight absent providers through the existing extension point:

| Provider | Handler |
|---|---|
| `Microsoft.Compute/virtualMachines` | `VmHandler` |
| `Microsoft.ContainerService/managedClusters` | `AksHandler` |
| `Microsoft.ContainerRegistry/registries` | `AcrHandler` |
| `Microsoft.Cache/Redis` | `RedisHandler` |
| `Microsoft.DBforPostgreSQL/flexibleServers` | `PostgresHandler` |
| `Microsoft.DBforMySQL/flexibleServers` | `MySqlHandler` |
| `Microsoft.DBforMariaDB/servers` | `MariaDbHandler` |
| `Microsoft.Sql/servers` | `SqlHandler` |

and then makes both listings answer from **one assembly**, `ArmHandler.indexedResources(sub, rg)`, where a null resource group means subscription scope — so they cannot disagree about the estate. `ResourceIndexContributor` gains `listSubscriptionResources(sub)` with no default implementation, so a service cannot appear in one listing while falling silently out of the other; Network and Managed Identity gain the subscription-scoped overload they lacked beside their resource-group one.

Each contributor answers from the state its own list endpoint already reads (`scanAll()` for the container-backed services, `state.listServersBy...` for the DB family), so the index cannot drift from the type-scoped listing. `ArmResources.indexEntry(...)` centralises the entry shape — `id`, `name`, `type`, `location`, `tags`, and no `properties`, which real Azure returns only under `$expand`. `AciModels` builds its entry through it too, so nine contributors share one definition rather than nine copies.

One deliberate deviation is preserved and now documented where the assembly happens: `ArmHandler`'s own Storage / Key Vault / Web state contributes its **full stored body**, not the trimmed entry, because the azurerm provider reads this listing to populate its Key Vault cache and looks vaults up by `properties.vaultUri`. A test pins it.

Widening the listing exposed that `$filter` was being ignored — and the azurerm provider fills its Key Vault cache from `$filter=resourceType eq 'Microsoft.KeyVault/vaults'`, parsing every returned id as a vault id ([populate_cache.go](https://github.com/hashicorp/terraform-provider-azurerm/blob/da2e1b2963ac19fb1ffdfac22adb05b1738708bc/internal/services/keyvault/client/populate_cache.go#L72)), so an unfiltered listing carrying a storage account failed `terraform destroy`. **`ArmResourceFilter`** now compiles the 2021-04-01 `$filter` contract — `eq`/`ne` on `resourceType`, `name`, `location`, `resourceGroup`, `tagName`, `tagValue`; `substringof(…)`; `startswith(tagName, …)`; `and`/`or` — and both listings apply it. A filter the emulator does not implement is answered `400 InvalidFilter`, never an unfiltered listing: returning more than was asked for is exactly what broke the client.

Cosmos, Service Bus, Event Hubs and App Configuration are deliberately untouched: they have no ARM management plane in floci-az, so they own no resource to index.

Closes #269.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

**Incorrect behavior:** real Azure returns every resource in scope from both listings. floci-az returned a subset from one and almost nothing from the other. A VM created in a group answered a direct `GET` with `200` and appeared under `/providers/Microsoft.Compute/virtualMachines`, yet the group listing came back without it — and the subscription listing came back empty on an estate with a storage account and a virtual network in it.

That resource-group listing is what the `azurerm` provider reads before deleting a resource group to verify it is empty; `ResourceIndexContributor`'s own javadoc records that a service which skips registration lets `terraform destroy` remove a group whose resources still exist. Both listings are also what anything enumerating an estate generically depends on — `az resource list`, the Resource Management SDKs, and drift checkers that diff deployed state against expected, which report real resources as missing.

Verified live against the release image and a build of this branch, same requests both times:

```
BEFORE — resource group holding a storage account and a VM
  rg index   -> [('reprosa', 'Microsoft.Storage/storageAccounts')]

BEFORE — subscription holding a storage account and a virtual network
  sub index  -> {"value":[]}

AFTER — two groups, each with a VM and a vnet; one also with a storage account
  rg-one     -> Microsoft.Storage/storageAccounts   subidxsa
                Microsoft.Network/virtualNetworks   vnet-rg-one
                Microsoft.Compute/virtualMachines   vm-rg-one

  subscription -> Microsoft.Storage/storageAccounts subidxsa
                  Microsoft.Network/virtualNetworks vnet-rg-one
                  Microsoft.Network/virtualNetworks vnet-rg-two
                  Microsoft.Compute/virtualMachines vm-rg-one
                  Microsoft.Compute/virtualMachines vm-rg-two
```

Tags round-trip into the index entry, `properties` stays out of it (matching `GenericResourceExpanded`), and the group listing stays scoped to its own group while the subscription listing spans every group.

The repo's own terraform compat, run locally with `|| true` taken off `destroy`: everything terraform had created was destroyed cleanly — Key Vault secret and vault included, with a storage account, VM and Redis cache in the same subscription — so azurerm's filtered cache query now gets only vaults. (`apply` itself stopped at ACR on my Mac because macOS AirPlay Receiver holds host port 5000, which the ACR sidecar needs; unrelated to this change.)

## Checklist

- [x] Tests pass locally — a bounded gate of **542 tests, 0 failures** on JDK 25, on the branch rebased onto 0.13.0: the whole `io.floci.az.core` package plus every test class in the packages this touches (`aci`, `acr`, `aks`, `vm`, `redis`, `postgres`, `mysql`, `mariadb`, `sql`, `arm`, `apim`, `managedidentity`, `network`, `containerapps`), excluding the `*DockerTest` classes. Full disclosure, same as in #253: the complete `./mvnw test` spins up docker-based service emulators I can't run in my sandbox, so I gated on those packages instead. CI ran the rest — the first commit came back green on all 13 checks including every SDK compat suite and compat-terraform/opentofu/azcli.
- [x] New or updated integration test added — one RestAssured case per contributing provider, in each service's existing mocked-mode test class, asserting **both** listings; a new `ArmResourceIndexTest` for what a per-service test cannot show: that the group listing stays scoped while the subscription listing spans groups, that a contributor's entry carries identity without `properties`, that a key vault's `vaultUri` survives, that azurerm's exact `$filter` query returns only vaults whatever else the estate holds, and that an unimplemented filter is refused; plus `ArmResourceFilterTest` over the grammar.
- Every assertion was proven to fire. Reverting `src/main` to the pre-fix state leaves 8 failures, one per newly-registered provider; reverting it to the resource-group-only state leaves 11, exactly the subscription-scope assertions; reverting only the `$filter` wiring leaves 3, exactly the filter assertions. With all three commits applied, all pass.

### DIFF
--- a/src/main/java/io/floci/az/core/arm/ArmResources.java
+++ b/src/main/java/io/floci/az/core/arm/ArmResources.java
@@ -15,6 +15,27 @@ public final class ArmResources {
     private ArmResources() {
     }
 
+    /**
+     * Minimal ARM resource entry for the resource-group {@code /resources} index —
+     * {@code id}, {@code name}, {@code type}, {@code location}, and {@code tags} when non-empty.
+     *
+     * <p>Azure's generic resource listing returns identity fields only; {@code properties} arrives
+     * solely under {@code $expand}. Contributors project their resource onto this shape rather than
+     * echoing the body their {@code GET} returns.</p>
+     */
+    public static Map<String, Object> indexEntry(String id, String name, String type, String location,
+                                                 Map<String, String> tags) {
+        Map<String, Object> entry = new LinkedHashMap<>();
+        entry.put("id", id);
+        entry.put("name", name);
+        entry.put("type", type);
+        entry.put("location", location);
+        if (tags != null && !tags.isEmpty()) {
+            entry.put("tags", tags);
+        }
+        return entry;
+    }
+
     /** Copy of the resource without the internal routing keys. */
     public static Map<String, Object> stripInternal(Map<String, Object> resource) {
         Map<String, Object> copy = new LinkedHashMap<>(resource);
--- a/src/main/java/io/floci/az/core/arm/ResourceIndexContributor.java
+++ b/src/main/java/io/floci/az/core/arm/ResourceIndexContributor.java
@@ -4,15 +4,21 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * A service that registers its resources in ARM's resource-group index —
- * {@code GET subscriptions/{sub}/resourceGroups/{rg}/resources}.
+ * A service that registers its resources in ARM's generic resource index — both
+ * {@code GET subscriptions/{sub}/resourceGroups/{rg}/resources} and
+ * {@code GET subscriptions/{sub}/resources}.
  *
- * <p>The azurerm provider calls that listing before deleting a resource group to verify it is
- * empty; a service that skips registration lets {@code terraform destroy} remove a group whose
- * resources still exist. {@code ArmHandler} discovers implementations via CDI
+ * <p>The azurerm provider calls the resource-group listing before deleting a resource group to
+ * verify it is empty; a service that skips registration lets {@code terraform destroy} remove a
+ * group whose resources still exist. Anything enumerating an estate generically rather than
+ * per-type — {@code az resource list}, the Resource Management SDKs, drift checkers — reads one
+ * of these two listings too. {@code ArmHandler} discovers implementations via CDI
  * {@code Instance<ResourceIndexContributor>} (the same pattern as {@code Resettable} /
  * {@code AdminController}), so contributing costs one interface — not an ArmHandler constructor
  * change per service.</p>
+ *
+ * <p>Both methods are abstract on purpose. A default returning nothing would let a new service
+ * fall out of one listing silently, which is the bug this interface exists to prevent.</p>
  */
 public interface ResourceIndexContributor {
 
@@ -21,6 +27,9 @@ public interface ResourceIndexContributor {
      * {@code tags}) for this service's resources in the given resource group.
      */
     List<Map<String, Object>> listRgResources(String sub, String rg);
+
+    /** The same maps for every one of this service's resources in the subscription. */
+    List<Map<String, Object>> listSubscriptionResources(String sub);
 
     /** Whether this contributor's service is enabled; disabled services contribute nothing. */
     default boolean indexEnabled() {
--- a/src/main/java/io/floci/az/services/aci/AciHandler.java
+++ b/src/main/java/io/floci/az/services/aci/AciHandler.java
@@ -513,9 +513,17 @@ public class AciHandler implements AzureServiceHandler, Resettable, ResourceInde
 
     @Override
     public List<Map<String, Object>> listRgResources(String sub, String rg) {
-        String prefix = (sub + "/" + rg + "/").toLowerCase();
+        return indexEntries((sub + "/" + rg + "/").toLowerCase());
+    }
+
+    @Override
+    public List<Map<String, Object>> listSubscriptionResources(String sub) {
+        return indexEntries((sub + "/").toLowerCase());
+    }
+
+    private List<Map<String, Object>> indexEntries(String storageKeyPrefix) {
         return scanAll().stream()
-                .filter(group -> group.storageKey().toLowerCase().startsWith(prefix))
+                .filter(group -> group.storageKey().toLowerCase().startsWith(storageKeyPrefix))
                 .map(ContainerGroup::indexEntry)
                 .toList();
     }
--- a/src/main/java/io/floci/az/services/aci/AciModels.java
+++ b/src/main/java/io/floci/az/services/aci/AciModels.java
@@ -2,10 +2,10 @@ package io.floci.az.services.aci;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import io.floci.az.core.arm.ArmResources;
 import io.quarkus.runtime.annotations.RegisterForReflection;
 
 import java.time.Instant;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -80,15 +80,8 @@ public class AciModels {
 
         /** Minimal ARM resource map for the resource-group {@code /resources} index. */
         public Map<String, Object> indexEntry() {
-            Map<String, Object> entry = new LinkedHashMap<>();
-            entry.put("id", armId());
-            entry.put("name", name);
-            entry.put("type", "Microsoft.ContainerInstance/containerGroups");
-            entry.put("location", location);
-            if (tags != null && !tags.isEmpty()) {
-                entry.put("tags", tags);
-            }
-            return entry;
+            return ArmResources.indexEntry(armId(), name,
+                    "Microsoft.ContainerInstance/containerGroups", location, tags);
         }
     }
 
--- a/src/main/java/io/floci/az/services/acr/AcrHandler.java
+++ b/src/main/java/io/floci/az/services/acr/AcrHandler.java
@@ -15,6 +15,8 @@ import io.floci.az.core.storage.StorageFactory;
 import io.floci.az.services.acr.AcrModels.Registry;
 import io.floci.az.core.arm.ArmErrors;
 import io.floci.az.core.arm.ArmPaths;
+import io.floci.az.core.arm.ArmResources;
+import io.floci.az.core.arm.ResourceIndexContributor;
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -62,13 +64,15 @@ import java.util.concurrent.TimeUnit;
  * and {@code loginServer} is the cosmetic {@code {name}.azurecr.io} for management-plane fidelity.</p>
  */
 @ApplicationScoped
-public class AcrHandler implements AzureServiceHandler, Resettable {
+public class AcrHandler implements AzureServiceHandler, Resettable, ResourceIndexContributor {
 
     private static final Logger LOG = Logger.getLogger(AcrHandler.class);
 
     private static final ObjectMapper MAPPER = new ObjectMapper()
             .registerModule(new JavaTimeModule())
             .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+
+    private static final String TYPE = "Microsoft.ContainerRegistry/registries";
 
     private static final SecureRandom RANDOM = new SecureRandom();
     private static final String ALNUM =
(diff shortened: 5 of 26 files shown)


hectorvent:
Thank you for extending this to the subscription scope; the estate-level test is the right shape. The widened listing breaks one real client, which I reproduced with the repo's own Terraform config.

**(blocking)** azurerm fills its Key Vault cache from `GET /subscriptions/{sub}/resources?$filter=resourceType eq 'Microsoft.KeyVault/vaults'` and parses every returned id as a vault id, failing hard on any other type ([populate_cache.go#L72](https://github.com/hashicorp/terraform-provider-azurerm/blob/da2e1b2963ac19fb1ffdfac22adb05b1738708bc/internal/services/keyvault/client/populate_cache.go#L72)). The handler ignores `$filter`, so the listing now carries the storage account and VM as well, and `terraform destroy` fails with `parsing ".../Microsoft.Storage/storageAccounts/flocitestsa" as a Key Vault ID`. Main destroys cleanly. CI missed it because the bats script runs destroy with `|| true`. Honouring `resourceType eq '...'` on both listings fixes it, and a test asserting the filtered result would pin it. The parameter is part of the listing's contract ([resources.json#L481](https://github.com/Azure/azure-rest-api-specs/blob/7efab4ca79331ad2ac48eb67fb944d030745918d/specification/resources/resource-manager/Microsoft.Resources/resources/stable/2021-04-01/resources.json#L481)).

**(follow-up, separate PR)** The `$expand` wording from my earlier note now also lives in the `indexedResources` javadoc.

MichaelDanCurtis:
Thanks — reproduced and confirmed at the source: `populateCache` sets `Filter = "resourceType eq 'Microsoft.KeyVault/vaults'"` on the `resources` list and runs `ParseKeyVaultIDInsensitively` over every id it gets back, so the widened listing was handing it a storage-account id.

Fixed in 9bba532, on top of a rebase onto 0.13.0:

- **`ArmResourceFilter`** compiles the 2021-04-01 `$filter` grammar the listings document — `eq`/`ne` on `resourceType`, `name`, `location`, `resourceGroup`, `tagName`, `tagValue`; `substringof(…)` and `startswith(tagName, …)`; `and`/`or` with `and` binding tighter; a `tagName`/`tagValue` pair read as one tag — and both listings apply it.
- A filter the emulator does not implement (`identity`, `plan`, parentheses) is answered **`400 InvalidFilter`** rather than an unfiltered listing. Returning more than was asked for is what broke the client, so ignoring is not the safe default.
- Pinned by `ArmResourceIndexTest.keyVaultCacheFilterReturnsOnlyVaults` — azurerm's exact query against an estate holding a vault, a storage account and two VMs returns only the vault, `vaultUri` intact — plus a resource-group-scope case, the 400 case, and `ArmResourceFilterTest` over the grammar. All three HTTP assertions fail with the wiring reverted.
- The rebase pulled in `managedHSMs` (folded into the same assembly) and `ContainerAppsHandler`, which already implemented `ResourceIndexContributor`; the abstract subscription method caught it at compile time, as intended, and it now answers both scopes.
- The `$expand` wording: corrected in both javadocs here rather than as a follow-up, since the text only exists on this branch — the generic listing never carries `properties`; `$expand` adds only `createdTime`, `changedTime` and `provisioningState`.

I ran the repo's terraform compat locally with the `|| true` taken off `destroy` (temporarily, in a scratch checkout). `apply` couldn't complete on my Mac — the ACR sidecar needs host port 5000 and macOS AirPlay Receiver holds it — but `destroy` of everything terraform had created went through cleanly, Key Vault secret and vault included, with the storage account, VM and Redis cache sitting in the same subscription. One thing worth showing from that run: azurerm then refused to delete the resource group because it still held `registries/flocitestacr` — the registry whose create had failed, so terraform never owned it but floci-az still had the record. That's the listing doing what it exists for; on CI, where ACR creates normally, the group empties.

I've left `|| true` in place: without a full clean run on this machine I can't prove a fatal `destroy` is green end-to-end, so removing it here would be gambling CI on paths I couldn't exercise. Happy to do it as a follow-up.

hectorvent:
Thank you for the quick and thorough fix, and for confirming it at the azurerm source. I repeated the same Terraform run that showed the failure, and `terraform destroy` now completes cleanly on 9bba532, Key Vault secret and vault included. Refusing unknown predicates with a 400 rather than returning everything is the right call, and the az CLI filter forms (`resourceGroup`, `tagname`, `startswith(tagname, ...)`) all parse too. Thanks as well for folding in the Container Apps method and the `$expand` wording.

**(follow-up, separate PR)** The `and`/`or` split does not respect quotes, so a tag value like `'dev and test'` is answered 400. Resource names cannot hold spaces, so only tag filters are affected.

Happy to take the `|| true` removal as a follow-up, as you offered. No blockers from my side.

hectorvent (approved):

PR #271, opened 2026-09-02
feat(containerapps): mocked-mode Container Apps, verified with a real terraform apply

johncgoat:
## Summary
- Adds mocked-mode ARM emulation for Azure Container Apps (`Microsoft.App`) — `managedEnvironments`, `containerApps`, and `jobs` — following the ACI/VM handler pattern (CDI auto-discovery, synchronous PUT/DELETE LRO shapes, Terraform-safe enum canonicalization/defaulting, secret hygiene, `ResourceIndexContributor` registration).
- Closes a gap the environment resource depends on: `MonitorHandler` now implements Log Analytics `sharedKeys` (deterministic, so repeated calls don't produce a Terraform diff) and the subscription-wide workspace list, both of which the azurerm provider calls during a Container App environment Create/Read.
- That fix required generalizing `ArmHandler.dispatch()` to delegate subscription-wide (`no /resourceGroups/`) provider-resource-type paths through the same `ArmProviderService` lane used for resource-group-scoped ones — which also surfaced and fixed a latent bug: `ApiManagementService`'s own subscription-wide list branch existed but was unreachable, and would have silently answered an always-empty `200` once reachable instead of the real list.
- Adds `azurerm_log_analytics_workspace` + `azurerm_container_app_environment` + `azurerm_container_app` to the `compat-terraform` smoke-test suite with bats spot checks — verified against a real `terraform apply`/`destroy` cycle (see Test plan).
- Fixes a routing-table golden-list test that was failing since the Container Apps provider route was never added to it, and closes doc/banner gaps (`docs/services/containerapps.md`, README/index service tables, startup banner) the earlier commits on this branch left open.

## Test plan
- [x] `./mvnw test` — 938 tests, all pass
- [x] Conventional Commits format on both commits
- [x] `CHANGELOG.md` not touched
- [x] Terraform verification (the point of this PR): ran `make test-terraform-compat` (Docker-based, matches CI's `compatibility.yml`) — `azurerm_log_analytics_workspace` → `azurerm_container_app_environment` → `azurerm_container_app` all created and destroyed cleanly with no errors. Two unrelated pre-existing resources in that same run (`azurerm_redis_cache`, `azurerm_container_registry` — real-Docker-sidecar-backed, neither touched by this PR) failed with `polling after Create: polling failed`, consistent with a local Docker-socket environment limitation on this machine rather than a regression; `compatibility.yml`'s CI run will confirm.
- [ ] Coverage: this repo has no coverage tooling configured (no JaCoCo, nothing in CI) — no gate to report.

## Docs
- App-context skill: this repo has no local Claude Code app-context skill (checked `~/.claude/skills/` for one matching "working in the floci-az repo"), so nothing to update there.
- `docs/services/containerapps.md` added; `docs/services/monitor.md`, `docs/services/index.md`, and `README.md` updated to cover the new endpoints and add the missing Container Apps row.


johncgoat:
Closing — opened in error, this work is staying on a private fork for internal use, not intended as an upstream contribution.

PR #276, opened 2026-09-04
fix(cosmos): evaluate array parameter membership

### LINKED ISSUE #274: [BUG] Cosmos DB: ARRAY_CONTAINS with an array parameter and document-field value returns incorrect results
## Service

Cosmos DB NoSQL SQL query engine.

## Summary

`ARRAY_CONTAINS(@arrayParameter, c.field)` does not match documents whose field is present in the supplied array. The negated form can consequently include documents that should be excluded.

## Environment

- Reproduced against `floci/floci-az:0.12.0` with .NET 10 and `Microsoft.Azure.Cosmos` 3.62.1.
- Gateway mode with `LimitToEndpoint = true`.
- Current query engine source at `a9e4799d29f86c1eb39a5bf2d2e72f8be380e20e` was also compiled and exercised in a standalone probe.

## Reproduction

Create a container with partition key `/pk` and these two documents:

```json
{"id":"alice","pk":"alice","type":"alice"}
{"id":"bob","pk":"bob","type":"bob"}
```

Run this query with the .NET SDK:

```csharp
var query = new QueryDefinition(
    "SELECT * FROM c WHERE ARRAY_CONTAINS(@types, c.type)")
    .WithParameter("@types", new[] { "alice" });

using var iterator = container.GetItemQueryIterator<dynamic>(
    query,
    requestOptions: new QueryRequestOptions
    {
        PartitionKey = new PartitionKey("alice")
    });

while (iterator.HasMoreResults)
{
    foreach (var item in await iterator.ReadNextAsync())
        Console.WriteLine((string)item.id);
}
```

Expected: `alice`.

Actual on 0.12.0: zero rows. The equivalent REST query also returned zero rows.

A standalone probe of the current query engine, supplied both documents, additionally showed:

| Predicate | Expected IDs | Actual IDs |
|---|---|---|
| `ARRAY_CONTAINS(@types, c.type)` | alice | none |
| `NOT ARRAY_CONTAINS(@types, c.type)` | bob | alice, bob |

Scalar equality and document-array/string-literal membership worked as control queries. The negated-result observation is from the standalone current-source probe; the positive case was also verified against the released container through the SDK.

## Source mechanism

The [`ARRAY_CONTAINS` evaluator](https://github.com/floci-io/floci-az/blob/a9e4799d29f86c1eb39a5bf2d2e72f8be380e20e/src/main/java/io/floci/az/services/cosmos/CosmosQueryEngine.java#L426) resolves its first argument as a document path and its second as a literal. That does not support an array-valued parameter as the first argument and a document-field expression as the second. [Parameter substitution](https://github.com/floci-io/floci-az/blob/a9e4799d29f86c1eb39a5bf2d2e72f8be380e20e/src/main/java/io/floci/az/services/cosmos/CosmosQueryEngine.java#L663) also stringifies array parameters.

## Impact

PalCal uses this query shape for notification kinds, event-ID lookup and mark-read candidate selection. It uses the negated shape for blocked/muted actor suppression. The current behavior can produce empty notification results or allow suppressed notifications through locally.

## Suggested acceptance

- Preserve array parameter values and evaluate both function arguments as expressions.
- Positive membership returns only matching documents; negated membership excludes them.
- Cover empty arrays, multiple values, numeric and string values, and strings containing commas or quotes without changing existing document-array membership behavior.
- Add real .NET SDK regression coverage.

## Related work / duplicate check

No matching issue or PR found after searching open and closed trackers. #162 / #168 cover correlated `EXISTS`, and #216 / #220 cover .NET query initialization. This is a separate expression-evaluation gap.

thomhurst:
## Summary

Cosmos queries using `ARRAY_CONTAINS(@ids, c.id)` returned no matches because array parameters lost their structure and membership arguments were treated as a field and a literal. Evaluate both arguments as expressions, preserve typed array/object values, and support negated membership through the existing predicate evaluator.

Parameter substitution now matches complete tokens outside quoted strings. Array values containing commas, apostrophes, repeated spaces or parameter-looking text retain their contents. General array/object expression evaluation also preserves the existing .NET ORDER BY wrapper support.

Closes #274

## Type of change

- [x] Bug fix (`fix:`)

## Azure Compatibility

Verified through Microsoft.Azure.Cosmos 3.62.0 in Gateway mode against the packaged JVM emulator. Added SDK assertions for included IDs, excluded IDs and empty arrays. Updated the supported-query documentation.

## Validation

- Four new regression tests failed before the fix and pass afterward.
- Focused query tests: 20 passed, including existing ORDER BY and EXISTS coverage.
- Full Java suite: 890 tests passed; Maven packaging succeeded.
- .NET Cosmos compatibility suites: 2 passed against this branch's packaged emulator, including transactional batch coverage.
- Full diff reviewed and `git diff --check` passed. Native image and other language SDK suites were not run.

## Checklist

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
@@ -57,6 +57,22 @@ public sealed class CosmosCompatibilityTests
             await Assert.That(selected.Select(item => item.Id))
                 .IsEquivalentTo(["one", "two", "three"]);
 
+            List<QueryItem> included = await ReadAll(
+                container.GetItemQueryIterator<QueryItem>(new QueryDefinition(
+                    "SELECT * FROM c WHERE ARRAY_CONTAINS(@ids, c.id)")
+                    .WithParameter("@ids", new[] { "one", "three" })), cancellationToken);
+            await Assert.That(included.Select(item => item.Id)).IsEquivalentTo(["one", "three"]);
+            List<QueryItem> excluded = await ReadAll(
+                container.GetItemQueryIterator<QueryItem>(new QueryDefinition(
+                    "SELECT * FROM c WHERE NOT ARRAY_CONTAINS(@ids, c.id)")
+                    .WithParameter("@ids", new[] { "one", "three" })), cancellationToken);
+            await Assert.That(excluded.Select(item => item.Id)).IsEquivalentTo(["two"]);
+            List<QueryItem> empty = await ReadAll(
+                container.GetItemQueryIterator<QueryItem>(new QueryDefinition(
+                    "SELECT * FROM c WHERE ARRAY_CONTAINS(@ids, c.id)")
+                    .WithParameter("@ids", Array.Empty<string>())), cancellationToken);
+            await Assert.That(empty.Count).IsEqualTo(0);
+
             List<QueryItem> ordered = await ReadAll(
                 container.GetItemQueryIterator<QueryItem>("SELECT * FROM c ORDER BY c.rank"),
                 cancellationToken);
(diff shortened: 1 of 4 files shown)


hectorvent (approved):
Thank you, this fixes more than the title says: the old substitution was substring based, so `@id` inside `@ids` was a latent bug as well, and the token-aware replacement closes it along with quoted text and whitespace inside literals. The .NET assertions for included, excluded and empty arrays run in the compat suite, so a real Microsoft.Azure.Cosmos client exercises the fix rather than only the engine tests.

One observation rather than a finding: the optional third `ARRAY_CONTAINS` argument (partial match) was ignored before and still is, so nothing regressed there. Worth a docs line, or a follow-up if anyone needs it.

No blockers from my side.

PR #279, opened 2026-09-04
fix: normalize Key Vault route trailing slashes

### LINKED ISSUE #238: [BUG] Requesting `/secrets/?....` should return same response as `/secrets?...`
## Short Explanation 

I was trying to explore Azure Key Vault with floci-az (dotnet c# client), it was working in `SecretClient` level, but when trying to add dotnet configuration from keyvault with 

```csharp 
   builder.Configuration.AddAzureKeyVault(secretClient, new AzureKeyVaultConfigurationOptions());
```
it fail to list all secret from `/secrets/` endpoint, as far as i understand, internally C# azure library will concat `/secrets/` with `?api-version=....`, it work without slash `/` but azure dotnet source code was use slash `/`

```csharp 
 // internally AddAzureKeyVault() is calling this 
 public Uri CreateFirstPageUri(string path)
    {
        RequestUriBuilder requestUriBuilder = new RequestUriBuilder();
        requestUriBuilder.Reset(VaultUri);
        requestUriBuilder.AppendPath(path, escape: false);
        requestUriBuilder.AppendQuery("api-version", ApiVersion);
        return requestUriBuilder.ToUri();
    }


public virtual Pageable<SecretProperties> GetPropertiesOfSecrets(CancellationToken cancellationToken = default(CancellationToken))
    {
        Uri firstPageUri = _pipeline.CreateFirstPageUri("/secrets/");
        return PageResponseEnumerator.CreateEnumerable((string? nextLink) => _pipeline.GetPage(firstPageUri, nextLink, () => new SecretProperties(), "SecretClient.GetPropertiesOfSecrets", cancellationToken));
    }

```

- **Secret Client Source** -> https://github.com/Azure/azure-sdk-for-net/blob/Azure.Security.KeyVault.Secrets_4.11.0/sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretClient.cs#L303
- **KeyVaultPipeLine** -> https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/Azure.Security.KeyVault.Shared/src/KeyVaultPipeline.cs

## Service

KeyVault 

## Expected behavior

requesting `/secrets/?....` should return same as `/secrets?....`

```shell
PS  > curl.exe -k `
>>   -H "Authorization: Bearer $token" `
>>   "https://localhost:4577/devstoreaccount1-keyvault/secrets?api-version=7.4"
{"value":[{"id":"https://devstoreaccount1.vault.azure.net/secrets/asdf","attributes":{"enabled":true,"nbf":null,"exp":null,"created":1787905426,"updated":1787905426,"recoveryLevel":"Purgeable","recoverableDays":7},"contentType":"text/plain","tags":{}}],"nextLink":null}
```

## Actual behavior

currently requesting `/secrets/?....` return error as follow

```shell
PS  > curl.exe -k `
>>   -H "Authorization: Bearer $token" `
>>   "https://localhost:4577/devstoreaccount1-keyvault/secrets/?api-version=7.4"
{"error":{"code":"SecretNotFound","message":"A secret with (name/id)  was not found in this key vault. If you recently deleted this secret, it may still be recoverable. For more information, see https://docs.microsoft.com/en-us/rest/api/keyvault/getsecret."}}
``` 

## Reproduction
- start floci-az 
- and make request with curl inside powershell as described above  

## Environment

- Floci-az version / image tag:
- Java SDK version (if applicable): 
```shell
PS  > java --version
openjdk 11.0.20.1 2023-08-24 LTS
OpenJDK Runtime Environment Microsoft-8297088 (build 11.0.20.1+1-LTS)
OpenJDK 64-Bit Server VM Microsoft-8297088 (build 11.0.20.1+1-LTS, mixed mode)
```
- How you're running Floci-az (Docker / native / `mvn quarkus:dev`): run with Docker

thomhurst:
## Summary

Key Vault treated `GET /secrets/` as a lookup for an empty secret name, returning `SecretNotFound` and breaking .NET `AddAzureKeyVault` configuration loading.

Normalize one optional trailing slash for every fixed-route equality check in `KeyVaultHandler`, covering secrets, deleted secrets, certificate contacts, and the root probe. Preserve the original path for named-resource parsing and leave other services' paths untouched.

Closes #238.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Azure.Security.KeyVault.Secrets 4.11.0 uses `/secrets/` and `/deletedsecrets/` collection paths. Added .NET coverage for empty and populated listing, synchronous and asynchronous enumeration, and `AddAzureKeyVault` configuration loading. HTTP regressions cover optional slashes, populated deleted-secret listings, authentication challenges, and method restrictions.

The .NET fixture sets explicit validity dates to isolate this routing regression from an existing, separate failure when the emulator emits null timestamp attributes. Timestamp serialization is addressed separately in #280.

## Validation

All relevant checks passed against the rebuilt emulator:

- `./mvnw package -Dtest=KeyVaultServiceTest,AzureRoutingFilterTest,DisabledServiceRoutingTest`: 38 tests.
- .NET Key Vault compatibility test: 1 test.
- Java Key Vault compatibility suite: 17 tests.
- Python Key Vault compatibility suite: 24 tests.
- JavaScript Key Vault compatibility suite: 14 tests.
- `git diff --check`.

Before the fix, the new HTTP regressions reproduced the 404 listing failure and incorrect acceptance of collection writes.

## Checklist

- [ ] `./mvnw test` passes locally (focused tests and relevant SDK suites run; full suite not run)

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/KeyVaultCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/KeyVaultCompatibilityTests.cs
@@ -1,0 +1,103 @@
+using Azure.Core;
+using Azure.Core.Pipeline;
+using Azure.Extensions.AspNetCore.Configuration.Secrets;
+using Azure.Security.KeyVault.Secrets;
+using Microsoft.Extensions.Configuration;
+
+namespace FlociAz.Compatibility;
+
+public sealed class KeyVaultCompatibilityTests
+{
+    private static readonly Uri EmulatorEndpoint = new(
+        Environment.GetEnvironmentVariable("FLOCI_AZ_ENDPOINT") ?? "http://localhost:4577");
+
+    [Test]
+    [Timeout(60_000)]
+    public async Task ListsSecretsAndLoadsConfiguration(CancellationToken cancellationToken)
+    {
+        using var httpClient = new HttpClient(new EmulatorHandler());
+        var vaultUri = new UriBuilder(EmulatorEndpoint)
+        {
+            Scheme = "https",
+            Path = $"/kv{Guid.NewGuid():N}-keyvault"
+        };
+        var client = new SecretClient(vaultUri.Uri, new FakeCredential(), new SecretClientOptions
+        {
+            DisableChallengeResourceVerification = true,
+            Transport = new HttpClientTransport(httpClient)
+        });
+
+        await Assert.That(client.GetPropertiesOfSecrets(cancellationToken).Any()).IsFalse();
+        // Explicit dates isolate routing from the emulator's existing null timestamp incompatibility.
+        await client.SetSecretAsync(new KeyVaultSecret("App--Message", "hello")
+        {
+            Properties =
+            {
+                NotBefore = DateTimeOffset.UtcNow.AddMinutes(-1),
+                ExpiresOn = DateTimeOffset.UtcNow.AddHours(1)
+            }
+        }, cancellationToken);
+
+        try
+        {
+            var names = client.GetPropertiesOfSecrets(cancellationToken)
+                .Select(secret => secret.Name).ToList();
+            await Assert.That(names).IsEquivalentTo(["App--Message"]);
+
+            var asyncNames = new List<string>();
+            await foreach (var secret in client.GetPropertiesOfSecretsAsync(cancellationToken))
+            {
+                asyncNames.Add(secret.Name);
+            }
+            await Assert.That(asyncNames).IsEquivalentTo(["App--Message"]);
+
+            using var configuration = new ConfigurationManager();
+            configuration.AddAzureKeyVault(client, new AzureKeyVaultConfigurationOptions());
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
Thank you, this is the right shape: normalise only the fixed-route comparisons and leave named-resource parsing on the original path. Your .NET test shows the released 4.11.0 client listing with `/secrets/`, and real Azure accepts that form, so accepting one trailing slash is the conservative reading (the regenerated 4.12 preview drops the slash, which is why newer clients never hit this). The `AddAzureKeyVault` case is the scenario from #238 and it runs in the compat suite.

Setting explicit dates in that fixture to sidestep the timestamp bug, and fixing that separately in #280, was the right call; the two branches merge cleanly against each other.

No blockers from my side.

PR #280, opened 2026-09-05
fix: omit unset Key Vault timestamps

thomhurst:
## Summary

Key Vault emitted `"nbf": null` and `"exp": null` for unset validity dates. Azure.Security.KeyVault.Secrets 4.11.0 reads present timestamp properties as integers, so ordinary `SetSecretAsync` calls failed with `InvalidOperationException` when either date was absent.

Omit unset validity dates from the shared attribute serializer. Preserve numeric values, including Unix epoch zero, and continue recognizing empty or `"null"` metadata produced by existing storage and update paths. This fixes secret bundles, list items, deleted secrets, and recovered secrets without changing persistence or request handling.

This is the independent timestamp follow-up discovered while fixing #238 in #279. Neither PR includes the other's implementation changes; the two branches merge cleanly.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

The [Azure Key Vault 7.4 attribute schema](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/data-plane/Legacy/stable/7.4/common.json) defines optional `nbf` and `exp` properties as integer Unix timestamps. The [.NET SDK attribute reader](https://github.com/Azure/azure-sdk-for-net/blob/Azure.Security.KeyVault.Secrets_4.11.0/sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretAttributes.cs) calls `GetInt64()` when either property is present.

Added automated .NET lifecycle coverage for neither date, either date individually, and both dates. HTTP regression tests verify actual key omission across CRUD, lists, updates, deletion, and recovery, plus preservation of numeric zero and handling of explicitly cleared dates.

## Validation

- Before the fix: all 6 new HTTP cases failed; 3 of 4 .NET timestamp combinations reproduced the null-to-number exception.
- After the fix: 27 focused Java tests passed (`KeyVaultTimestampTest`, `AzureRoutingFilterTest`, `DisabledServiceRoutingTest`).
- .NET Key Vault timestamp suite: 4 passed.
- Existing Java / Python / JavaScript Key Vault suites: 17 / 24 / 14 passed.
- `./mvnw package -DskipTests` passed after stopping the local test emulator to release Windows JAR locks.
- `git diff --check` passed.

## Checklist

- [ ] `./mvnw test` passes locally (focused tests and relevant SDK suites run; full suite not run)

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/KeyVaultTimestampCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/KeyVaultTimestampCompatibilityTests.cs
@@ -1,0 +1,114 @@
+using Azure.Core;
+using Azure.Core.Pipeline;
+using Azure.Security.KeyVault.Secrets;
+
+namespace FlociAz.Compatibility;
+
+public sealed class KeyVaultTimestampCompatibilityTests
+{
+    private static readonly Uri EmulatorEndpoint = new(
+        Environment.GetEnvironmentVariable("FLOCI_AZ_ENDPOINT") ?? "http://localhost:4577");
+
+    [Test]
+    [Arguments(false, false)]
+    [Arguments(true, false)]
+    [Arguments(false, true)]
+    [Arguments(true, true)]
+    [Timeout(60_000)]
+    public async Task OptionalTimestampsSurviveSecretLifecycle(
+        bool includeNotBefore, bool includeExpiresOn, CancellationToken cancellationToken)
+    {
+        using var httpClient = new HttpClient(new EmulatorHandler());
+        var vaultUri = new UriBuilder(EmulatorEndpoint)
+        {
+            Scheme = "https",
+            Path = $"/kv{Guid.NewGuid():N}-keyvault"
+        };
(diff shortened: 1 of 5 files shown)


hectorvent (approved):
Thank you, a small change with the right blast radius: one serializer, every response shape. The 7.4 schema declares `nbf` and `exp` as optional integers ([common.json#L16](https://github.com/Azure/azure-rest-api-specs/blob/7efab4ca79331ad2ac48eb67fb944d030745918d/specification/keyvault/data-plane/Legacy/stable/7.4/common.json#L16)) and the .NET reader calls `GetInt64()` whenever the property is present ([SecretAttributes.cs#L48](https://github.com/Azure/azure-sdk-for-net/blob/e345636942bb5b24c090a6df28e6b4bf65f8f8e4/sdk/keyvault/Azure.Security.KeyVault.Secrets/src/SecretAttributes.cs#L48)), so `null` was a genuine wire deviation rather than a client quirk. Keeping epoch zero numeric and treating a cleared date as absent covers the two edges I would have asked about.

No blockers from my side.

PR #281, opened 2026-09-05
feat(policy): add Azure Policy control plane (definitions, set definitions, assignments, exemptions)

Tal-E:
## Summary

Adds the Azure Policy control plane under `Microsoft.Authorization`: policy definitions, policy set definitions (initiatives), policy assignments and policy exemptions, at every ARM scope the real service accepts (management group, subscription, resource group, resource; tenant-rooted built-in reads answer empty). This lets `az policy ...`, the `azurerm` policy resources and the policy SDKs (`armpolicy`, `azure-mgmt-resource`, `azure-resourcemanager-resources`) author and deploy policies against floci-az instead of a real subscription.

Scope is deliberately the control plane only: rules are stored and validated structurally, never evaluated. Nothing is denied, audited or remediated, there is no compliance state (`Microsoft.PolicyInsights` is not implemented) and no built-in definitions are seeded. The docs page lists every deviation.

What is in the change:

- `services/policy/`: `PolicyHandler` (filter-lane `AzureServiceHandler`, guarded provider route on `Microsoft.Authorization` so role assignments, locks and deny assignments keep falling through to the generic ARM handler), `PolicyStore` (in-memory, keyed by lower-cased resource id like the other ARM control-plane stores) and `PolicyPath` (scope / type / name parsing for extension resources).
- Definitions and set definitions: CreateOrUpdate, Get, Delete, List with `$filter=policyType eq '...'`. Server-side `policyType=Custom`, `mode` default `Indexed`, `version` default `1.0.0`, `metadata.createdBy/createdOn/updatedBy/updatedOn` stamps, generated `policyDefinitionReferenceId` and `definitionVersion=1.*.*` on references.
- Assignments: Create, Get, Update (PATCH of identity, location, resourceSelectors, overrides), Delete (returns the deleted assignment), List for subscription / resource group / resource / management group with `atScope()`, `atExactScope()`, `atScopeAndBelow()` and `policyDefinitionId eq '...'`. `scope`, `enforcementMode`, `definitionVersion`, `instanceId` populated server-side; `SystemAssigned` identities get a stable principal, `UserAssigned` identities resolve `principalId`/`clientId` from the Managed Identity store when the identity exists.
- Exemptions: CreateOrUpdate, Get, Update (PATCH), Delete (returns the body), List with `atScope()`, `atExactScope()`, `excludeExpired()`, `policyAssignmentId eq '...'`; deleted together with their assignment.
- Referential integrity: scoped definition / set definition references must exist (`PolicyDefinitionNotFound`, `PolicySetDefinitionNotFound`), exemptions need an existing assignment (`PolicyAssignmentNotFound`), and referenced definitions / set definitions cannot be deleted (`InvalidDeletePolicyDefinitionRequest`, `InvalidDeletePolicySetDefinitionRequest`). Built-in ids (`/providers/Microsoft.Authorization/policyDefinitions/{guid}`) are accepted without lookup.
- Routing: `routeArmProviders` now also considers tenant-rooted `providers/Microsoft.X/...` paths (matching the `/providers/<ns>/` marker against the slash-prefixed path), which is how built-in listings and management-group scopes arrive. Subscription-rooted behaviour is unchanged; `RoutingTableAssemblyTest` gains the new golden entry and the guarded-first invariant now covers both guarded routes.
- Config `floci-az.services.policy.enabled` (`FLOCI_AZ_SERVICES_POLICY_ENABLED`, default `true`), banner line, `docs/services/policy.md`, nav, service matrix and README rows.
- Tests: `PolicyTest` (definitions, set definitions, assignments, exemptions, scope filters, management-group and tenant scopes, referential integrity, routing boundaries), `PolicyDisabledTest`, two new cases in `AzureRoutingFilterTest`; a Java compat test (`PolicyCompatibilityTest`, raw REST like the Managed Identity one) and an `az policy` BATS file for the az CLI suite.

Out of scope, left for follow-ups: rule evaluation / deny effects, PolicyInsights compliance and remediation, built-in definition catalogue, the `versions` sub-resource API, pagination of listings.

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

- Request/response shapes follow `azure-rest-api-specs` policy `stable/2025-03-01/openapi.json` and exemptions `preview/2022-07-01-preview/policyExemptions.json`; any `api-version` is accepted.
- Response conventions were cross-checked against the Azure CLI's recorded live traffic in `Azure/azure-cli` (`test_resource_policy_default.yaml`, `test_resource_policyset_default.yaml`, `test_resource_policyexemption_default.yaml`, api-version 2025-11-01): `201` on definition and assignment PUT, `200` on set definition / exemption update, `metadata.createdBy/createdOn` stamps, generated numeric `policyDefinitionReferenceId`, `definitionVersion` `1.*.*`, assignment `instanceId`, DELETE of assignments and exemptions returning the resource body, and the exact `PolicyDefinitionNotFound` / `PolicySetDefinitionNotFound` messages.
- `InvalidDeletePolicyDefinitionRequest` (definition referenced by a set definition) reproduces the live message reported in hashicorp/terraform-provider-azurerm#15615.
- Not verified against a live tenant: the `InvalidCreatePolicyDefinitionRequest` / `InvalidCreatePolicyAssignmentRequest` / `InvalidCreatePolicyExemptionRequest` codes returned for missing required properties (modelled on the recorded `InvalidCreatePolicySetDefinitionRequest`), and the assignment-referenced variant of `InvalidDeletePolicyDefinitionRequest`. Happy to adjust if a maintainer has live samples.
- Verification SDKs: the Java compat test drives the REST wire protocol directly (java.net.http, as `ManagedIdentityCompatibilityTest` does); the BATS file uses `az` from `mcr.microsoft.com/azure-cli:latest`.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <!-- 
 AI Context: This is Floci-Az, a lightweight Local Azure Emulator. 
 Identity: It is the Azure equivalent of Floci (AWS). It is NOT LocalStack.
-Protocols: Implements Azure Storage (Blob, Queue, Table), Azure Functions, App Configuration, Key Vault, Event Hubs, Service Bus (Microsoft.ServiceBus), Cosmos DB, Azure SQL Database, Azure Database for PostgreSQL (Microsoft.DBforPostgreSQL), Azure Kubernetes Service (AKS), API Management (Microsoft.ApiManagement), Virtual Network (Microsoft.Network), Virtual Machines (Microsoft.Compute), Azure Cache for Redis (Microsoft.Cache), Azure Container Registry (Microsoft.ContainerRegistry), Azure Container Instances (Microsoft.ContainerInstance), Event Grid (Microsoft.EventGrid), Azure Monitor / Log Analytics (Microsoft.OperationalInsights / Microsoft.Insights), Communication Services Email (Microsoft.Communication), Managed Identity (Microsoft.ManagedIdentity + IMDS token endpoint), Microsoft Entra ID (OpenID Connect / OAuth2 token issuance, including interactive auth-code+PKCE sign-in), and a narrow Microsoft Graph slice (service principal discovery, group membership).
+Protocols: Implements Azure Storage (Blob, Queue, Table), Azure Functions, App Configuration, Key Vault, Event Hubs, Service Bus (Microsoft.ServiceBus), Cosmos DB, Azure SQL Database, Azure Database for PostgreSQL (Microsoft.DBforPostgreSQL), Azure Kubernetes Service (AKS), API Management (Microsoft.ApiManagement), Virtual Network (Microsoft.Network), Virtual Machines (Microsoft.Compute), Azure Cache for Redis (Microsoft.Cache), Azure Container Registry (Microsoft.ContainerRegistry), Azure Container Instances (Microsoft.ContainerInstance), Event Grid (Microsoft.EventGrid), Azure Monitor / Log Analytics (Microsoft.OperationalInsights / Microsoft.Insights), Communication Services Email (Microsoft.Communication), Managed Identity (Microsoft.ManagedIdentity + IMDS token endpoint), Azure Policy (Microsoft.Authorization policy definitions, set definitions, assignments and exemptions), Microsoft Entra ID (OpenID Connect / OAuth2 token issuance, including interactive auth-code+PKCE sign-in), and a narrow Microsoft Graph slice (service principal discovery, group membership).
 Default Port: 4577 (HTTP; also HTTPS when FLOCI_AZ_TLS_ENABLED=true via protocol-sniffing proxy). AMQP port: 5672 (Event Hubs). Kafka port: 9093 (Event Hubs, opt-in). k3s API: 6443-7443 (AKS). Redis: 6379-6399 (Azure Cache for Redis).
 Tech Stack: Java, Quarkus, Docker-in-Docker for Functions. Artemis sidecar for Event Hubs AMQP. Redpanda sidecar for Kafka. k3s sidecar for AKS. Redis sidecar for Azure Cache for Redis.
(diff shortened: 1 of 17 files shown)


hectorvent:
Thank you, this is a well-scoped control plane and the description is a model: CLI cassettes named per claim, unverified codes listed, deviations on the docs page. Handler-owned routing, `Resettable`, constructor injection and the Managed Identity store pattern are all followed.

**(blocking)** Your own `az policy exemption create` case fails in CI: the CLI exits with status 3 (its not-found class) while the other seventeen policy cases pass. The suite image is `mcr.microsoft.com/azure-cli:latest`, whose generated exemption command PUTs `/{scope}/providers/Microsoft.Authorization/policyExemptions/{name}` at api-version 2026-01-01-preview with the scope built from `-g`, so an older local CLI may not show it. Reproducing with that image and capturing stderr should point at the assignment id lookup or the newer body shape.

On Greptile: the 2025-03-01 assignment identity enum is `SystemAssigned`, `UserAssigned`, `None` with no combined type ([openapi.json#L4275](https://github.com/Azure/azure-rest-api-specs/blob/7efab4ca79331ad2ac48eb67fb944d030745918d/specification/resources/resource-manager/Microsoft.Authorization/policy/stable/2025-03-01/openapi.json#L4275)), so `None` is correct; the documented resource scope includes `resourceGroups`; the non-atomic maps are the pattern every ARM store here uses.

**(follow-up, separate PR)** Narrow the built-in bypass to `/providers/Microsoft.Authorization/` ids, and add a Java `PolicyClient` test beside the raw-REST one.

That is the only blocker.

hectorvent:
Two things since my review. A rebase is now owed, and the cause is my merge batch: #268 landed and both branches add tests at the end of `AzureRoutingFilterTest`, so main conflicts there. And the exemption case above still stands. Rebasing onto main first also gives you the current suite to reproduce it against.

PR #285, opened 2026-09-07
fix(eventhub): accept the latest position unquoted, as .NET writes it

deyanp:
The follow-up you asked for on #242, now that it has landed — based on `main`, so nothing is stacked this time.

## The bug

The SDKs disagree about quoting the filter operand. Java and Rust send `amqp.annotation.x-opt-offset > '@latest'`; .NET builds it into the expression bare ([`AmqpFilter.cs#L62`](https://github.com/Azure/azure-sdk-for-net/blob/e345636942bb5b24c090a6df28e6b4bf65f8f8e4/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpFilter.cs#L62), your pointer). Only the quoted form was recognised, so a .NET consumer opening at the latest position got:

```
amqp.annotation.x-opt-offset > @latest   ->   floci_offset > @latest
```

`@latest` is not a valid selector token, so this is refused at parse time — the attach fails outright rather than merely matching nothing.

## Scope, which is narrower than it first looks

I probed `EventHubFilterSupport` directly rather than reasoning about the regex:

```
amqp.annotation.x-opt-offset > '@latest'             ->  floci_enqueued_time >= 1788352451343
amqp.annotation.x-opt-offset > @latest               ->  floci_offset > @latest        <-- the bug
amqp.annotation.x-opt-offset > -1                    ->  floci_offset > -1
amqp.annotation.x-opt-sequence-number >= 12345       ->  floci_sequence_number >= 12345
amqp.annotation.x-opt-enqueued-time > 1756400000000  ->  floci_enqueued_time > 1756400000000
```

Only `@latest` breaks. The numeric positions were never affected: the name mapping does not care about quoting, and the unquoting step only has to fire when the SDK supplied quotes — so .NET's offset, sequence-number and enqueued-time positions already worked.

## The change

One pattern accepts both spellings:

```java
- "amqp\\.annotation\\.x-opt-offset\\s*>=?\\s*'@latest'"
+ "amqp\\.annotation\\.x-opt-offset\\s*>=?\\s*(?:'@latest'|@latest)"
```

An explicit alternation rather than `'?@latest'?`, which would also accept a mismatched quote — no SDK sends that, and quietly accepting malformed input is not a feature.

## Testing

`EventHubFilterSupport` had no test of its own, which is exactly how one client's spelling came to stand in for all of them. It has one now, covering each form an SDK emits — quoted and unquoted, across all three annotations, plus selectors it must leave alone.

I checked the test actually bites: reverting the pattern fails it on precisely the .NET form and nothing else.

```
1007 tests, 0 failures
```

(`AksDockerTest` and `PostgresDockerTest` are excluded — they fail the same way on a clean `main` on this machine, which runs k3d and a dozen containers.)

### DIFF
--- a/src/artemis-amqp-patch/java/org/apache/activemq/artemis/protocol/amqp/proton/EventHubFilterSupport.java
+++ b/src/artemis-amqp-patch/java/org/apache/activemq/artemis/protocol/amqp/proton/EventHubFilterSupport.java
@@ -16,9 +16,16 @@ public final class EventHubFilterSupport {
 
    private static final String ANNOTATION_PREFIX = "amqp.annotation.";
 
-   /** {@code amqp.annotation.x-opt-offset > '@latest'} — the start position with no numeric form. */
+   /**
+    * {@code amqp.annotation.x-opt-offset > '@latest'} — the start position with no numeric form.
+    *
+    * <p>The quotes are optional because the SDKs disagree about them: the Java and Rust clients
+    * quote the operand, the .NET one does not. Accepting either costs nothing, and requiring them
+    * refused a .NET consumer's attach outright — {@code @latest} is no more a valid selector token
+    * unquoted than quoted, so what reached the parser had nothing where the operand should be.
+    */
    private static final Pattern LATEST =
-      Pattern.compile("amqp\\.annotation\\.x-opt-offset\\s*>=?\\s*'@latest'");
+      Pattern.compile("amqp\\.annotation\\.x-opt-offset\\s*>=?\\s*(?:'@latest'|@latest)");
 
    /** A quoted operand on one of the rewritten properties, e.g. {@code floci_offset > '-1'}. */
    private static final Pattern QUOTED_OPERAND = Pattern.compile(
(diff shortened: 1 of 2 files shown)


hectorvent:
Thank you, this is the follow-up done properly: on main rather than stacked, with the claim sourced and the gap that allowed it closed. I checked the .NET client and it does build the filter unquoted ([AmqpFilter.cs#L62](https://github.com/Azure/azure-sdk-for-net/blob/e345636942bb5b24c090a6df28e6b4bf65f8f8e4/sdk/eventhub/Azure.Messaging.EventHubs/src/Amqp/AmqpFilter.cs#L62)), so the old pattern refused a .NET attach outright rather than merely matching nothing. Probing every position form before scoping the change is what makes this reviewable, and that `EventHubFilterSupport` had no test of its own is exactly how one client's spelling came to stand in for all of them.

**(follow-up, separate PR)** The test hardcodes the patch jar name including `2.44.0`. #282 moved that version into the pom alone so the jar names, the container paths and the image tag cannot drift apart, and `ArtemisPatchVersions.ARTEMIS_AMQP_PATCH_RESOURCE` gives you the same string. As written, the next Artemis bump breaks this test with a file-not-found rather than the guard's message.

No blockers from my side.

hectorvent (approved):

Issue #194, opened 2026-08-13
[FEAT]: Add Azure Key Vault support (CRUD + Crypto ops) and Managed HSM


parthcodes:
## Service

Azure Key Vault (`Microsoft.KeyVault`)

## API Action / Feature

Key Vault Keys object model (CRUD + versions + deleted keys), Cryptographic Operations (Encrypt/Decrypt, Sign/Verify, Wrap/Unwrap), `GetRandomBytes`, Key Rotation Policy, and Managed HSM (`Microsoft.KeyVault/managedHSMs`).

## Azure Documentation

- [Key Vault REST API Reference](https://learn.microsoft.com/en-us/rest/api/keyvault/)
- [Keys Data Plane Operations](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/)
- [Key Cryptographic Operations (Encrypt)](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/encrypt/encrypt)
- [Managed HSM Management REST API](https://learn.microsoft.com/en-us/rest/api/keyvault/mhsm/managed-hsms)

## Why is this needed?

`floci-az`'s Key Vault emulation is currently **secrets-only**. The entire **Keys** side of Azure Key Vault — both the object model and the cryptographic operations — is missing, along with **Managed HSM**. 

As a result, applications or infrastructure tools (Terraform, Azure SDKs, CLI) attempting key management or cryptographic workflows fail when targeting `floci-az`:
- **Envelope encryption / data protection at rest** (`KeyClient` + `CryptographyClient`) fail on key creation/fetch or `encrypt`/`decrypt`/`wrapKey`/`unwrapKey` ops.
- **JWT / Token signing & verification** workflows fail on `sign` / `verify`.
- **HSM-backed key management** workflows targeting Managed HSM (`.managedhsm.azure.net`) 404.

### Code Investigation & Root Cause

1. **Handler Dispatch Fallthrough**:
   All Key Vault data-plane logic lives in `src/main/java/io/floci/az/services/keyvault/KeyVaultHandler.java`. The `handle()` method dispatch (`lines 70-123`) routes `secrets`, `deletedsecrets`, and `certificates/contacts`, but has no `keys/` or `deletedkeys/` branches. Any key request (e.g. `PUT /{account}-keyvault/keys/mykey`) falls through to `404 Resource not found` (`line 122`).

2. **Routing Filter Configuration**:
[...]

PR #287, opened 2026-09-08
feat:  Adding Azure Key Vault keys and cryptography

parthcodes:
## Summary

Adds Azure Key Vault **keys** and **cryptography** to floci-az (as requested in #194 ), plus a Managed HSM flavor, and fixes the self-signed TLS certificate so strict X.509 validators (e.g. Rust's rustls-platform-verifier) can trust floci-az's HTTPS endpoint.

Two parts:

1. **`feat(keyvault)`** — key CRUD (RSA/EC/oct, including `-HSM` variants), versions, soft-delete/recover/purge, backup/restore, key rotation + rotation policy, encrypt/decrypt/sign/verify/wrap/unwrap (`RSA-OAEP`, `RSA-OAEP-256`, `RSA1_5`, AES-GCM; `RS*`/`PS*`/`ES*`), and `/rng`. The same data plane is served under the `/{account}-managedhsm/` suffix, with ARM `Microsoft.KeyVault/managedHSMs` CRUD.
2. **`fix(tls)`** — the self-signed HTTPS certificate was a single, CA-flagged certificate presented in the leaf/end-entity position, which strict path validators reject (`CaUsedAsEndEntity`). floci-az now generates a proper CA → leaf chain and serves the CA at `GET /_floci/tls-cert` as the trust anchor. This also closes a SAN gap by adding `*.managedhsm.azure.net`.

## Type of change

- [x] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

- **Keys / crypto / Managed HSM** verified with the Azure SDKs for Java (`azure-security-keyvault-keys` + cryptography), Python (`azure-keyvault-keys` ≥ 4.7.0), JavaScript (`@azure/keyvault-keys` ^4.9.0), and the `az keyvault key` CLI (crypto ops, rotation policy, soft-delete lifecycle). Wire protocol follows the Key Vault REST API (`api-version=7.4`); intentional deviations are documented in `docs/services/key-vault.md`.
- **TLS fix** — previous behavior: a CA-flagged cert was presented as the server leaf, so strict validators (rustls) failed the handshake. New behavior: a CA-signed, non-CA leaf chain; verified end-to-end with `openssl verify` and `openssl s_client -verify_hostname` returning `Verify return code: 0` against the CA served at `/_floci/tls-cert`.

## Checklist


hectorvent:
Thank you for this, and sorry about the conflict: merging #228 changed `AzureRoutingFilter`, the same method your account suffix hardening edits. That one file is the only conflict, everything else auto merges.

**(blocking)** Your CI has run now, and `RSA-OAEP-256` fails in the Python, Node and az CLI suites. `Cipher.getInstance("RSA/ECB/OAEPWithSHA-256AndMGF1Padding")` without an `OAEPParameterSpec` gives digest SHA-256 with MGF1 **SHA-1** under SunJCE (reproduced on Java 25: `MD: SHA-256, MGF: MGF1SHA-1`). Clients use MGF1 SHA-256 ([rsa_encryption.py#L67](https://github.com/Azure/azure-sdk-for-python/blob/c9e6a7495525298ec64e5451a2f7422447027415/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/rsa_encryption.py#L67)) and encrypt locally, so only the decrypt reaches floci-az and the padding check fails. An explicit `new OAEPParameterSpec("SHA-256", "MGF1", MGF1ParameterSpec.SHA256, PSource.PSpecified.DEFAULT)` fixes it.

Worth saying: the Java suite is green because the Java SDK has no local OAEP-256 path ([RsaOaep.java#L64](https://github.com/Azure/azure-sdk-for-java/blob/e2eddb6460286bb37546d75b0aeac0eb78a24d89/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/cryptography/implementation/RsaOaep.java#L64) is SHA-1 only), so both halves of that round trip go through the emulator and agree. Easy thing to be caught by.

The rebase and that padding change unblock this.

parthcodes:
@hectorvent  
Both items are addressed:
1. RSA-OAEP-256 now uses an explicit OAEPParameterSpec("SHA-256", "MGF1", MGF1ParameterSpec.SHA256, PSource.PSpecified.DEFAULT) for encrypt/decrypt (wrap/unwrap share that path), plus an interop test that encrypts with MGF1 SHA-256 and decrypts through floci-az.

2. Rebased onto the AzureRoutingFilter changes from earlier PR #228, so that conflict is gone.

hectorvent:
Thank you, and this is cleared. CI has now run the full matrix on this head and all twelve checks pass, including the node suite, which was the last thing either of us could not prove.

Worth recording what the two fixes are, because both were found the hard way. The MGF1 pin is right: `RSA-OAEP-256` now passes an explicit `OAEPParameterSpec` for both the digest and the mask generation function, which is what SunJCE will not infer from the transformation string alone. Your test earns its keep too, encrypting with a separate cipher before calling `/decrypt`, so it would fail against the old code rather than agreeing with it.

The host-style kid plus transport rewrite is now the same pattern in both the Python and node suites, which is the right outcome.

No blockers from my side.

PR #293, opened 2026-09-09
fix(blob): allow write SAS append blocks

snazy:
Summary:
- Allow write-only SAS tokens to append blocks to existing Append Blobs.
- Preserve append-only SAS authorization and existing append conditions and lease checks.

Type of change:
- [x] Bug fix (fix:)

Azure Compatibility:
- Matches Azurite Append Block authorization, which accepts either a or w permission.

Checklist:
- [x] ./mvnw test passes locally

### DIFF
--- a/src/main/java/io/floci/az/core/auth/StorageSasAuthorization.java
+++ b/src/main/java/io/floci/az/core/auth/StorageSasAuthorization.java
@@ -396,7 +396,7 @@ public class StorageSasAuthorization {
         APPEND {
             @Override
             boolean allowedBy(StorageSasToken token) {
-                return token.hasPermission('a');
+                return token.hasAnyPermission('a', 'w');
             }
         };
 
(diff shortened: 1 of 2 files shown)


hectorvent:
Thank you, and the Azurite basis you cite checks out. `AppendBlob_AppendBlock` is declared as [Add plus Write](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/authentication/OperationBlobSASPermission.ts#L249) in both the blob scoped and container scoped tables, and the matcher returns true on the first declared character the token carries, so either permission authorizes the operation. `hasAnyPermission('a', 'w')` is the right shape, and it matches what `Operation.CREATE` next door already does.

Nice touch putting the new test beside the append only one from #236, so the pair now covers both permissions rather than replacing one case with the other.

No blockers from my side.

hectorvent (approved):

PR #294, opened 2026-09-09
test: derive the patched jar name from the pom, not a literal

deyanp:
The follow-up you asked for on #285.

## The problem

Three tests open the patched `artemis-amqp-protocol` jar through a loader of their own, and each spelled the file name out with the Artemis version in it:

```java
Path patchJar = Path.of("target", "classes", "artemis",
        "artemis-amqp-protocol-2.44.0-floci-az-artemis-amqp-patch.jar");
```

#282 moved that version into the pom alone, so the jar names, the container paths and the sidecar image tag cannot drift apart. A literal in a test puts it back in a second place — and the next Artemis bump would fail these three with a file-not-found instead of the message `ArtemisPatchVersionsTest` exists to give.

## The change

One helper, `PatchJarLoader`, derives the name from `ArtemisPatchVersions.ARTEMIS_AMQP_PATCH_RESOURCE` — a resource path, so it resolves against `target/classes` once the leading slash is gone. All three tests call it.

A helper rather than three copies of the derivation, because the literal spread the same way: copied from one test into the next as each new one was written. One spelling is now the only spelling — `grep -rn "2.44.0" src/test/` finds nothing.

It also reports a missing jar as a missing jar, naming the build step that writes it, rather than leaving a `NoSuchFileException` from a URL nobody can see.

## Scope

`ServiceBusDeadLetterSupportTest` predates the Event Hubs work and carried the same literal. I fixed it too — leaving it would have left the trap in place for exactly the bump that trips it, and it is the test the other two were modelled on.

## Testing

```
1037 tests, 0 failures
```

(`AksDockerTest` and `PostgresDockerTest` are excluded — they fail the same way on a clean `main` on this machine, which runs k3d and a dozen containers.)

One thing I could not verify end to end, so I would rather say it than imply otherwise: I tried simulating a bump by setting `artemis.version` to a fictitious `2.45.0`, but the build then fails at dependency resolution before any test runs, so it proves nothing about the test path. The guarantee rests on the derivation itself plus the four tests passing against the real version — including `ArtemisPatchVersionsTest`, which is the guard that should speak first when a bump forgets the image.

### DIFF
--- a/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AmqpEntityAddressTest.java
+++ b/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AmqpEntityAddressTest.java
@@ -6,9 +6,7 @@ import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
 import java.lang.reflect.Method;
-import java.net.URL;
 import java.net.URLClassLoader;
-import java.nio.file.Path;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
@@ -24,10 +22,7 @@ class AmqpEntityAddressTest {
 
     @BeforeAll
     static void loadFromPatchJar() throws Exception {
-        Path patchJar = Path.of("target", "classes", "artemis",
-                "artemis-amqp-protocol-2.44.0-floci-az-artemis-amqp-patch.jar");
-        loader = new URLClassLoader(new URL[]{patchJar.toUri().toURL()},
-                AmqpEntityAddressTest.class.getClassLoader());
+        loader = PatchJarLoader.open();
         toEntityPath = Class.forName(
                         "org.apache.activemq.artemis.protocol.amqp.proton.AmqpEntityAddress",
                         true, loader)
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you, this is the follow-up from #285 and it closes it properly, including the Service Bus test that carried the same literal from before the Event Hubs work.

On the part you could not verify: the derivation itself checks out. The constant keeps its leading slash, so dropping it and resolving against `target/classes` lands exactly on the `destfile` that the `package-servicebus-artemis-patches` step writes, and the execution id you quote in the error message is the real one. Saying plainly what you could not prove, rather than implying it was covered, is the right call.

The bot's point about the class initializer running first is true, but I do not think it needs fixing. `ArtemisPatchVersions` already fails with a message naming that same build step, so either path tells the reader what to run.

No blockers from my side.

hectorvent (approved):

deyanp:
thanks @hectorvent, that closes the last of it for me. When will you create a release?

The Event Hubs work is all merged now (#237, #241, #242, #246, #285, #294), but the latest release is 0.12.0 from 1 September, which predates #242 and #246. I want to point the CI at floci-az for Event Hubs, and to do that I'd rather pin a released tag than :nightly, which moves under us, or a dated nightly-mmddyyyy, which reads as a temporary pin nobody remembers to update.

hectorvent:
Hi @deyanp, thank you for your contribution. 
Releases are already scheduled, they are first and third Tuesday each month. More here https://github.com/floci-io/floci-az#release-train

PR #295, opened 2026-09-10
feat(blob): support snapshot-scoped user delegation SAS

snazy:
## Summary
- Persist immutable Blob snapshots and route snapshot reads and deletes to the captured state.
- Enforce sr=bs against the signed snapshot timestamp so snapshot SAS tokens cannot be used for the base blob.

## Type of change
- [x] New feature (feat:)

## Azure Compatibility
- Matches Blob Snapshot and user-delegation SAS snapshot-resource semantics.

## Checklist

### DIFF
--- a/src/main/java/io/floci/az/services/blob/BlobLeaseService.java
+++ b/src/main/java/io/floci/az/services/blob/BlobLeaseService.java
@@ -269,6 +269,15 @@ public class BlobLeaseService {
         return null;
     }
 
+    /**
+     * Snapshot Blob ignores an active lease unless the caller explicitly supplies a lease ID.
+     * When supplied, the ID must still identify the active lease. Call only inside
+     * {@link #exclusively}.
+     */
+    Response validateSnapshot(AzureRequest request, String blobKey) {
+        return header(request, "x-ms-lease-id") == null ? null : validateWrite(request, blobKey);
+    }
+
     /** Data Lake variant of {@link #validateWrite} using the DFS JSON error envelope. */
     Response validateDataLakeWrite(AzureRequest request, String blobKey) {
         String requestLeaseId = header(request, "x-ms-lease-id");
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you, this is careful work, and the lease and delete semantics land exactly where Azurite puts them.

First, on the open bot finding: `authorizeCreate` is right, please leave it. Azurite declares `Blob_CreateSnapshot` as [Create plus Write](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/authentication/OperationBlobSASPermission.ts#L169) in both tables and matches any one of them, so a `c` only SAS legitimately authorizes a snapshot.

**(blocking)** `SnapshotOperationNotSupported` is not a Blob error code. The close names that do exist are `PreviousSnapshotOperationNotSupported` and the Files one, which is probably where it came from. A mutation aimed at `?snapshot=` is [400 SnapshotsPresent](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/errors/StorageErrorFactory.ts#L416), not 409.

**(blocking)** The immutability guard sits below the `PUT ?comp=lease` branch, so `?comp=lease&snapshot=<ts>` still leases the base blob.

**(follow-up, separate PR)** `docs/services/blob.md:133` still says snapshots are not modeled.

snazy:
@greptile-apps I think, Azure explicitly permits Snapshot Blob with either c or w for a user-delegation SAS. Changing snapshotBlob to require w would reduce Azure compatibility. Please re-assess.

snazy:
Review-findings addressed in this PR and prepared #312 for the follow-up

hectorvent:
Thanks for the fast fix. Both points are closed at e9ce2ab: the immutability guard now sits above the lease branch, so a lease aimed at a snapshot is rejected, and the error is SnapshotsPresent at 400 with the test extended to pin the lease case too. Carrying the docs update in #312 works well.

No blockers from my side.

hectorvent (approved):

PR #304, opened 2026-09-13
fix(tls): allow non-localhost Host headers over HTTPS

parthcodes:
## Summary

Fixes TLS-enabled floci-az rejecting every non-localhost `Host` header with an HTTP 400 (empty body), which broke host-based routing such as `{account}.vault.azure.net` and `{account}.managedhsm.azure.net` over HTTPS. With TLS disabled the same requests returned 200.

Root cause: `TlsConfigSource` binds `quarkus.http.host=127.0.0.1`, and Quarkus auto-installs `HostValidationFilter` (restricted to `localhost` / `127.0.0.1` / `[::1]`) whenever it binds a loopback address with no explicit host-validation config. The filter runs at the Vert.x router stage, upstream of JAX-RS.

## Type of change

- [x] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Incorrect behavior: `GET /_floci/health` and every Key Vault data-plane call with a non-localhost `Host` header returned `400` when `FLOCI_AZ_TLS_ENABLED=true`, while the identical requests returned `200` with TLS disabled.

Fix: set `quarkus.http.host-validation.require-localhost=false` on the TLS-enabled path only, so the loopback auto-install no longer applies. `quarkus.http.host-validation.allowed-hosts` is deliberately left unset: it is evaluated before `require-localhost` and is an exact-match check with no wildcard support, so it would break both unbounded account names and `localhost`. The TLS-disabled path (which binds `0.0.0.0`) is left untouched.

## Verification

- `./mvnw test` green: 1154 tests, 0 failures, 0 errors.
- `make test-azcli` green: all 4 Key Vault data-plane tests that were previously skip-guarded now pass over HTTPS (`secret set/show`, `key create/show/list/delete`, `encrypt/decrypt + rotation-policy`, `soft-delete lifecycle`).

## Checklist

- [x] New or updated integration test added (2 unit tests pin the Quarkus-internal property; un-skipped the azcli Key Vault data-plane tests that now exercise this path)


hectorvent:
Thank you for this, and for the root cause writeup rather than just the one line fix. The detail that `HostValidationFilter` is auto installed only when Quarkus binds a loopback address, and that it runs at the Vert.x router stage upstream of JAX-RS, is exactly what makes this hard to find from the symptom.

The paragraph explaining why `allowed-hosts` is the wrong lever is the part I would have asked about otherwise. Evaluated before `require-localhost`, exact match, no wildcards, so it would break both unbounded account names and `localhost` itself.

I also went through both bot comments and neither stands: the port 443 requirement they ask you to document is already in your docs hunk, and the test they call environment dependent sets the property explicitly rather than clearing it.

No blockers from my side.

hectorvent (approved):

PR #305, opened 2026-09-13
fix(cosmos): respect NOT precedence in compound predicates

### LINKED ISSUE #300: [BUG] Cosmos DB: NOT precedence excludes valid rows from compound predicates
## Summary

The Cosmos SQL evaluator gives a leading `NOT` lower precedence than the following `OR`/`AND` expression. Valid predicates used to include non-deleted documents instead exclude them.

Reproduced over live HTTP against a JVM build of upstream `main` at `efdedff5cf2956f28f2f7baede3a40ff0a7a301a` on 2026-09-12. This was a build of that commit, not the published 0.12.0 image, although the Maven project still reports version 0.12.0.

## Reproduction

Create a database and a container partitioned by `/userId`. Create four documents with this shape, using IDs `item-1` through `item-4`:

```json
{"id":"item-1","userId":"user-1","isDeleted":false,"isActive":true,"claimedUntil":null}
```

POST the following query to the container's `/docs` endpoint with `Content-Type: application/query+json`, `x-ms-documentdb-isquery: true`, and `x-ms-documentdb-partitionkey: ["user-1"]`:

```json
{"query":"SELECT VALUE c.id FROM c WHERE (NOT IS_DEFINED(c.isDeleted) OR c.isDeleted = false)"}
```

**Expected:** all four IDs. `NOT` binds to `IS_DEFINED(c.isDeleted)`, and the second operand of `OR` is true.

**Actual:** `Documents: []`, `_count: 0`.

The logically equivalent query below returns all four IDs, isolating the precedence error:

```json
{"query":"SELECT VALUE c.id FROM c WHERE ((NOT IS_DEFINED(c.isDeleted)) OR c.isDeleted = false)"}
```

The same four documents also incorrectly produce zero results for:

```sql
SELECT VALUE c.id FROM c
WHERE (NOT IS_DEFINED(c.claimedUntil) OR IS_NULL(c.claimedUntil))
```

## Mechanism

[`CosmosQueryEngine.evalExpr`](https://github.com/floci-io/floci-az/blob/efdedff5cf2956f28f2f7baede3a40ff0a7a301a/src/main/java/io/floci/az/services/cosmos/CosmosQueryEngine.java#L280) processes a leading `NOT` before splitting top-level `OR` and `AND`. It therefore interprets `NOT A OR B` as `NOT (A OR B)`.

[Cosmos operator precedence](https://learn.microsoft.com/en-us/cosmos-db/query/operators) is `NOT`, then `AND`, then `OR`, from highest to lowest priority. Recursive evaluation must respect those binding rules.

## Application impact

PalCal uses the affected non-deleted predicate in notification lists/counts and active device-token queries. Its notification outbox also uses the nullable `claimedUntil` predicate. The failure can hide valid notifications and prevent eligible work from being selected.

Application references:

- [Notification queries](https://github.com/pal-cal/PalCal/blob/55f68e44f9ce36bfa63e4eca197bc4bc0bf2a6a1/src/PalCal.Notifications/Services/Cosmos/NotificationsCosmosRepository.NotificationQueries.cs#L107)
- [Outbox eligibility](https://github.com/pal-cal/PalCal/blob/55f68e44f9ce36bfa63e4eca197bc4bc0bf2a6a1/src/PalCal.Notifications/Services/Cosmos/NotificationsCosmosRepository.Outbox.cs#L124)

## Suggested acceptance

- Add failing-first regression coverage for `NOT A OR B`, `NOT A AND B`, nested parentheses, and explicit `NOT (A OR B)`.
- Verify the two application predicates above return the expected rows, including absent, null, false, and true property cases as appropriate.
- Exercise the predicate through a .NET Cosmos SDK query in the compatibility suite.

The live reproduction used HTTP requests, not the full PalCal application or its .NET integration suite. Reviewed open Cosmos PRs #296 and #297; neither addresses this precedence mechanism.

thomhurst:
Cosmos predicates such as `NOT IS_DEFINED(c.isDeleted) OR c.isDeleted = false` currently negate the whole expression and hide valid rows. Split top-level OR and AND before evaluating NOT, preserving parentheses and the documented logical precedence. Keep the AND inside BETWEEN ranges separate from logical conjunctions.

Adds exhaustive three-operand truth tables, correlated EXISTS regressions, negated BETWEEN cases, and .NET SDK coverage for non-deleted documents, absent/null claim timestamps, and range predicates.

Validation of the latest changes:
- Cosmos unit/integration tests: 177 passed.
- Live .NET Cosmos SDK tests: 10 passed.
- Live Java Cosmos SDK tests: 31 passed, including the explicitly enabled embedded NoSQL suite.
- Live Node.js Cosmos and Queue Storage SDK tests: 29 passed.
- Live Python Cosmos SDK tests: 19 passed.
- JVM package build and git diff --check passed.

The previous Node.js CI failure was the Queue Storage visibility-timeout test; the Cosmos suite passed in that run. Both suites pass locally on the latest changes.

Fixes #300.

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosCompatibilityTests.cs
@@ -141,6 +141,66 @@ public sealed class CosmosCompatibilityTests
         }
     }
 
+    [Test]
+    [Arguments("NOT IS_DEFINED(c.isDeleted) OR c.isDeleted = false", "absent,available,inactive,claimed")]
+    [Arguments("NOT IS_DEFINED(c.isDeleted) AND c.isActive = true", "absent")]
+    [Arguments("((NOT IS_DEFINED(c.isDeleted)) OR c.isDeleted = false)", "absent,available,inactive,claimed")]
+    [Arguments("NOT (IS_DEFINED(c.isDeleted) OR c.isActive = true)", "")]
+    [Arguments("((NOT IS_DEFINED(c.isDeleted) OR c.isDeleted = false) AND c.isActive = true)", "absent,available,claimed")]
+    [Arguments("NOT IS_DEFINED(c.claimedUntil) OR IS_NULL(c.claimedUntil)", "absent,available,deleted,inactive")]
+    [Timeout(60_000)]
+    public async Task DotnetSdkRespectsNotPrecedence(
+        string predicate, string expectedIds, CancellationToken cancellationToken)
+    {
+        using var client = new CosmosClient(
+            $"{EmulatorEndpoint}/devstoreaccount1-cosmos/",
+            CosmosKey,
+            new CosmosClientOptions
+            {
+                ConnectionMode = ConnectionMode.Gateway,
+                LimitToEndpoint = true
+            });
+        Database database = await client.CreateDatabaseAsync(
+            $"dotnet-not-{Guid.NewGuid():N}", cancellationToken: cancellationToken);
+
+        try
+        {
+            Container container = await database.CreateContainerAsync(
+                new ContainerProperties("items", "/userId"), cancellationToken: cancellationToken);
+            Dictionary<string, object?>[] documents =
+            [
+                new() { ["id"] = "absent", ["userId"] = "user-1", ["isActive"] = true },
+                new() { ["id"] = "available", ["userId"] = "user-1", ["isActive"] = true,
+                    ["isDeleted"] = false, ["claimedUntil"] = null },
+                new() { ["id"] = "deleted", ["userId"] = "user-1", ["isActive"] = true,
+                    ["isDeleted"] = true, ["claimedUntil"] = null },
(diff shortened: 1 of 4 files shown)


hectorvent:
Thank you, and the precedence fix is the right shape: splitting the top level `OR` and `AND` before applying `NOT` is what makes `NOT IS_DEFINED(c.isDeleted) OR c.isDeleted = false` behave, and the three operand truth tables are a good way to pin it.

I checked the bot's `BETWEEN` P1 rather than take either side on trust. It was a real defect when it was filed, and your later commit closes it: the scanner now recognises a depth zero `BETWEEN` and consumes exactly one following `AND` as a range separator, and the flag resets so a second `AND` still splits as a conjunction.

One honest note on my end. I could not settle this against a reference, because no local Azure source carries the Cosmos NoSQL grammar or its operator precedence. Your truth tables and the SDK runs are what the behaviour rests on here.

No blockers from my side.

hectorvent (approved):

PR #306, opened 2026-09-13
fix(cosmos): preserve query continuation after page deletion

### LINKED ISSUE #301: [BUG] Cosmos DB: continuation tokens skip remaining documents after consumed pages are deleted
## Summary

Cosmos query continuation tokens store a numeric offset into a newly evaluated result set. Deleting documents from an already consumed page shifts the remaining results before that offset, so continuation silently skips documents and can report completion while matching documents remain.

Reproduced over live HTTP against a JVM build of upstream `main` at `efdedff5cf2956f28f2f7baede3a40ff0a7a301a` on 2026-09-12. This was a build of that commit, not the published 0.12.0 image, although the Maven project still reports version 0.12.0.

## Reproduction

1. Create a database and a container partitioned by `/userId`.
2. Insert four documents with IDs `item-1` through `item-4`, each with `userId: "user-1"`.
3. POST this query to the container's `/docs` endpoint:

   ```json
   {"query":"SELECT c.id FROM c"}
   ```

   Use these headers:

   ```http
   Content-Type: application/query+json
   x-ms-documentdb-isquery: true
   x-ms-documentdb-partitionkey: ["user-1"]
   x-ms-max-item-count: 2
   ```

4. Save the response's `x-ms-continuation` token and delete the two returned documents using the same partition header.
5. Repeat the identical query with the saved `x-ms-continuation` header.
6. Repeat the query once more without a continuation header to inspect the remaining documents.

Observed in the live run:

```json
{
  "deleted": ["item-1", "item-2"],
  "continuedCount": 0,
  "remaining": ["item-3", "item-4"]
}
```

**Expected:** resuming after the first page still visits the remaining two documents. Deleting already consumed rows should not advance past unconsumed rows.

**Actual:** the continuation response is empty and has no next token, while a fresh query still finds two documents.

## Mechanism

[`CosmosHandler.queryDocuments`](https://github.com/floci-io/floci-az/blob/efdedff5cf2956f28f2f7baede3a40ff0a7a301a/src/main/java/io/floci/az/services/cosmos/CosmosHandler.java#L1190) re-evaluates all live matching documents, decodes `skip` from the token, then applies `subList(skip, ...)`. The token is only base64-encoded JSON containing that offset.

After deleting the first two of four rows, the fresh result contains two rows. Applying `skip: 2` removes both from the continuation response.

[Cosmos pagination documentation](https://learn.microsoft.com/en-us/cosmos-db/query/pagination) describes continuation tokens as bookmarks for query progress. This report concerns deleting already consumed results, not a request for snapshot isolation over arbitrary concurrent writes.

## Application impact

[PalCal's email purge](https://github.com/pal-cal/PalCal/blob/55f68e44f9ce36bfa63e4eca197bc4bc0bf2a6a1/src/PalCal.Email/Services/Cosmos/EmailCosmosRepository.cs#L366) reads a page of sent-log IDs, deletes those IDs through transactional batches, and then calls `ReadNextAsync` on the same iterator. When the query spans pages, the emulator's offset mechanism can leave sent logs behind while the purge loop reports completion.

The reproduction deliberately used a page size of two. The full PalCal/.NET purge path and its default page sizing were not executed in this investigation.

## Suggested acceptance

- Add a failing-first regression that reads a page, deletes its documents, and resumes until completion; every originally remaining document must be visited.
- Cover both unordered queries and `ORDER BY` queries, with partition scope preserved.
- Use a continuation representation that preserves progress through unconsumed results rather than a positional offset into a shrinking result set.
- Add a .NET Cosmos SDK compatibility case that deletes each received page before requesting the next, including transactional-batch deletes.

Reviewed open Cosmos PRs #296 and #297; neither addresses query continuation after deletion.

thomhurst:
Deleting documents from a consumed query page shifts the old numeric offset, causing continuation to skip documents that still exist. Use a stateless bookmark containing the source document identity and ORDER BY values, with a stable tie-breaker, before projecting each page. Preserve partition scope and total TOP/OFFSET/LIMIT bounds.

Legacy tokens retain their original offset ordering until completion. Decode integral offsets as 64-bit values, rejecting fractional, negative, and overflowing values without truncation. Bind new tokens to query text, parameters, account, container identity, and partition scope so accidental cross-query or cross-resource reuse returns 400. Scope normalization accepts equivalent parameter ordering, nested object member ordering, and numeric representations while preserving array order. Missing stored document identities now produce an explicit diagnostic, and the .NET purge test bounds its pagination loop more tightly.

Adds HTTP regression coverage, a Java SDK continuation case, and .NET SDK tests that delete each page through transactional batches before continuing.

Validation of the latest changes:
- Cosmos unit/integration tests: 193 passed.
- Live .NET Cosmos SDK tests: 8 passed, including page deletion, TOP/OFFSET/LIMIT bounds, and iterator recreation.
- Live Java Cosmos SDK tests: 32 passed, including the explicitly enabled embedded NoSQL suite.
- Live Node.js Cosmos SDK tests: 19 passed.
- Live Python Cosmos SDK tests: 19 passed.
- JVM package build and git diff --check passed.

Documented limits: aggregates, GROUP BY, DISTINCT, and legacy tokens retain offset pagination. Bookmarks do not provide snapshot isolation for concurrent inserts or changes to sort values.

Fixes #301.

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
@@ -11,6 +11,84 @@ public sealed class CosmosTransactionalBatchCompatibilityTests
         Environment.GetEnvironmentVariable("FLOCI_AZ_ENDPOINT") ?? "http://localhost:4577";
 
     [Test]
+    [Arguments(false)]
+    [Arguments(true)]
+    [Timeout(60_000)]
+    public async Task QueryContinuesAfterDeletingEachPage(bool ordered, CancellationToken cancellationToken)
+    {
+        using CosmosClient client = CreateClient($"dotnetpurge{Guid.NewGuid():N}");
+        Database database = await client.CreateDatabaseAsync(
+            $"db-{Guid.NewGuid():N}", cancellationToken: cancellationToken);
+        try
+        {
+            Container container = await database.CreateContainerAsync(
+                new ContainerProperties("items", "/tenant"), cancellationToken: cancellationToken);
+            string[] expected = Enumerable.Range(0, 7).Select(i => $"item-{i}").ToArray();
+            for (int i = 0; i < expected.Length; i++)
+            {
+                await container.CreateItemAsync(new { id = expected[i], tenant = "target", rank = i / 3 },
+                    new PartitionKey("target"), cancellationToken: cancellationToken);
+                await container.CreateItemAsync(new { id = expected[i], tenant = "other", rank = i / 3 },
+                    new PartitionKey("other"), cancellationToken: cancellationToken);
+            }
+
+            string sql = "SELECT c.id FROM c" + (ordered ? " ORDER BY c.rank DESC" : "");
+            var options = new QueryRequestOptions { PartitionKey = new PartitionKey("target"), MaxItemCount = 2 };
+            using FeedIterator<JObject> iterator = container.GetItemQueryIterator<JObject>(sql, requestOptions: options);
+            var visited = new List<string>();
+            int pages = 0;
+            while (iterator.HasMoreResults)
+            {
+                FeedResponse<JObject> page = await iterator.ReadNextAsync(cancellationToken);
+                if (++pages > 7)
+                {
+                    throw new InvalidOperationException("Query continuation did not terminate");
+                }
+                if (page.Count == 0)
+                {
+                    continue;
+                }
+                await Assert.That(page.Count <= 2).IsTrue();
+                TransactionalBatch batch = container.CreateTransactionalBatch(new PartitionKey("target"));
+                foreach (JObject item in page)
+                {
+                    string id = item.Value<string>("id")!;
+                    visited.Add(id);
+                    batch.DeleteItem(id);
+                }
+                using TransactionalBatchResponse deleted = await batch.ExecuteAsync(cancellationToken);
+                await Assert.That(deleted.IsSuccessStatusCode).IsTrue();
+                await Assert.That(deleted.Count).IsEqualTo(page.Count);
+                for (int i = 0; i < deleted.Count; i++)
+                {
+                    await Assert.That(deleted[i].StatusCode).IsEqualTo(HttpStatusCode.NoContent);
+                }
+            }
+
+            await Assert.That(visited).IsEquivalentTo(expected);
+            using FeedIterator<JObject> remaining = container.GetItemQueryIterator<JObject>(
+                "SELECT c.id FROM c", requestOptions: options);
+            while (remaining.HasMoreResults)
+            {
+                await Assert.That((await remaining.ReadNextAsync(cancellationToken)).Count).IsEqualTo(0);
+            }
+            using FeedIterator<JObject> other = container.GetItemQueryIterator<JObject>(
+                "SELECT c.id FROM c",
+                requestOptions: new QueryRequestOptions { PartitionKey = new PartitionKey("other") });
+            var preserved = new List<string>();
+            while (other.HasMoreResults)
+            {
+                preserved.AddRange((await other.ReadNextAsync(cancellationToken)).Select(item => item.Value<string>("id")!));
+            }
+            await Assert.That(preserved).IsEquivalentTo(expected);
+        }
+        finally
+        {
+            await database.DeleteAsync(cancellationToken: cancellationToken);
+        }
+    }
+
+    [Test]
     [Timeout(60_000)]
     public async Task DotnetSdkExecutesTransactionalBatches(CancellationToken cancellationToken)
     {
--- a/src/main/java/io/floci/az/services/cosmos/CosmosHandler.java
+++ b/src/main/java/io/floci/az/services/cosmos/CosmosHandler.java
@@ -1186,26 +1186,21 @@ public class CosmosHandler implements AzureServiceHandler, Resettable {
         } catch (IllegalArgumentException e) {
             return errorResponse(400, "BadRequest", e.getMessage());
         }
-        CosmosQueryEngine.QueryResult result = queryEngine.execute(parsed, scopedDocs);
-
-        // ---- Pagination ----
         int maxItemCount = parseMaxItemCount(req.headers().getHeaderString("x-ms-max-item-count"));
-        int skip         = decodeContinuationToken(req.headers().getHeaderString("x-ms-continuation"));
-
-        List<Object> allItems  = result.items();
-        List<Object> pageItems = skip > 0 ? allItems.subList(Math.min(skip, allItems.size()), allItems.size())
-                                          : new ArrayList<>(allItems);
-
-        String nextToken = null;
-        if (maxItemCount > 0 && pageItems.size() > maxItemCount) {
-            nextToken = encodeContinuationToken(skip + maxItemCount);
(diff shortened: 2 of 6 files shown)


src/main/java/io/floci/az/services/cosmos/CosmosHandler.java
```diff
@@ -1243,23 +1238,49 @@ private int parseMaxItemCount(String header) {
...
+            return null;
+        }
         try {
             String json = new String(Base64.getDecoder().decode(token), StandardCharsets.UTF_8);
             Map<?, ?> map = MAPPER.readValue(json, Map.class);
-            return ((Number) map.get("skip")).intValue();
-        } catch (Exception e) {
-            return 0;
+            if (map == null) {
+                throw new IllegalArgumentException("Invalid continuation bookmark");
```
utafrali, line 1292: Jackson maps JSON integers to `Integer` only when the value fits within int range; larger values come back as `Long`. Under `DeserializationFeature.USE_LONG_FOR_INTS` all JSON integers are `Long`. Both cases make `!(skip instanceof Integer consumed)` throw, turning a valid legacy continuation into a spurious 400. The previous `((Number) map.get("skip")).intValue()` was deliberately type-agnostic; this is a regression. Suggested fix:
```java
if (!(skip instanceof Number n) || n.longValue() < 0) {
    throw new IllegalArgumentException("Invalid continuation offset");
}
int consumed = n.intValue();
```

src/main/java/io/floci/az/services/cosmos/CosmosQueryEngine.java
```diff
@@ -122,6 +126,66 @@ public QueryResult execute(ParsedQuery q, List<Map<String, Object>> documents) {
...
+                          QueryContinuation continuation, int maxItemCount) {
+        int consumed = continuation == null ? 0 : continuation.consumed();
+        // Legacy tokens must keep their original ordering, including ties, until the query completes.
+        boolean legacyOffset = continuation != null && continuation.rid() == null;
+        if (legacyOffset || q.countQuery() || q.aggregateType() != null || !q.groupBy().isEmpty() || q.distinct()) {
+            List<Object> items = execute(q, documents).items();
+            int start = Math.min(consumed, items.size());
+            int size = pageSize(items.size() - start, maxItemCount);
+            List<Object> page = items.subList(start, start + size);
+            QueryContinuation next = start + size < items.size()
```
utafrali, line 142: `thenComparing(doc -> (String) doc.get("_rid"))` will throw `NullPointerException` during the sort if any document is missing `_rid`, because the `Comparator.thenComparing` overload that takes a `Function` calls `key1.compareTo(key2)` with no null guard on the extracted key. The same problem recurs in `compareContinuation` at `((String) doc.get("_rid")).compareTo(continuation.rid())`. If `_rid` is guaranteed non-null by the document store, an `Objects.requireNonNull` assertion at the top of `executePage` turns any violation into a clear failure message instead of a cryptic NPE mid-sort.

compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
```diff
@@ -10,6 +10,84 @@ public sealed class CosmosTransactionalBatchCompatibilityTests
...
+            var visited = new List<string>();
+            int pages = 0;
+            while (iterator.HasMoreResults)
+            {
+                FeedResponse<JObject> page = await iterator.ReadNextAsync(cancellationToken);
+                if (++pages > 7)
+                {
+                    throw new InvalidOperationException("Query continuation did not terminate");
+                }
+                if (page.Count == 0)
```
utafrali, line 47: With 7 items at `MaxItemCount = 2` the expected page count is exactly 4 (pages of 2, 2, 2, 1). The guard `pages > 7` lets the test run 3 extra rounds before bailing on an infinite-loop regression, which means potentially 3 extra network calls wasted in CI before the test times out. `pages > 5` would catch the same bug and exit much sooner. The `Assert.That(visited).IsEquivalentTo(expected)` check already verifies correctness; this guard just needs to be a practical early-exit.

utafrali (approved):
Solid fix that replaces fragile numeric-offset continuation with a stable document-bookmark approach. The core `executePage` logic correctly handles ORDER BY tie-breaking via `_rid`, preserves TOP/OFFSET/LIMIT bounds across pages, and keeps legacy tokens on their original offset path. Test coverage spans unit, HTTP, Java SDK, and .NET SDK layers. Three minor issues worth addressing before merge: a type-narrowing regression in the legacy token decoder, an unguarded `_rid` cast that can produce a misleading NPE, and an overly lenient loop guard in the .NET regression test.

hectorvent:
Thank you, and thank you [@utafrali](https://github.com/utafrali) for the review, which did the heavy lifting here. I went through all three of your points at the current head rather than assume either way. Every one was real when you raised it, and every one is now closed: the offset decode is a `JsonNode` integral check kept as a `long`, which also rejects fractional values, and the `_rid` guard is the `Objects.requireNonNull` you suggested, sitting immediately before the comparator. The bot's separate scope-binding point is closed too.

**(follow-up, separate PR)** The new Java continuation case lands in `CosmosNoSqlEngineCompatibilityTest`, which skips itself unless the NoSQL engine is enabled, and CI never sets that variable, so the green run here does not actually cover it. That is our gap rather than the PR's, and enabling the engine explicitly in the local run was the right call.

No blockers from my side.

hectorvent (approved):

src/main/java/io/floci/az/services/cosmos/CosmosHandler.java
thomhurst: Addressed in 33cc5709434fc5128dcdf06eb17db7b364140af6, already included in the current PR head. The decoder now reads a JsonNode, accepts integral values that fit in a nonnegative long, and retains long offsets through pagination. This avoids both the Integer/Long restriction and truncation through intValue(). HTTP regressions cover 2147483648, 4294967296, and Long.MAX_VALUE, plus rejection of fractional, negative, overflowing, and string offsets. Re-ran CosmosQueryContinuationTest and CosmosQueryEngineContinuationTest: all 39 tests passed, with no skips.

src/main/java/io/floci/az/services/cosmos/CosmosQueryEngine.java
thomhurst: Addressed in 33cc5709434fc5128dcdf06eb17db7b364140af6, already included in the current PR head. Bookmark pagination now checks every source document with Objects.requireNonNull before sorting, with the diagnostic "Stored Cosmos document is missing _rid required for query continuation". Added missingStoredIdentityHasClearDiagnostic to verify the failure message. Re-ran both continuation test classes: all 39 tests passed, with no skips.

compatibility-tests/sdk-test-dotnet/CosmosTransactionalBatchCompatibilityTests.cs
thomhurst: Addressed in 33cc5709434fc5128dcdf06eb17db7b364140af6, already included in the current PR head. Tightened the purge test guard from ++pages > 7 to ++pages > 5 while retaining the exact visited-item and partition-isolation assertions. The existing validation log for the current head records all 8 live .NET Cosmos SDK tests passing, with no skips, including both ordered and unordered page-deletion cases.

PR #307, opened 2026-09-13
fix(servicebus): include broker metadata on received messages

### LINKED ISSUE #302: [BUG] Service Bus: receives return zero SequenceNumber and minimum EnqueuedTime
## Service

Azure Service Bus: AMQP receives from queues, session-enabled topic subscriptions, and their dead-letter subqueues.

## Summary

Messages received with `Azure.Messaging.ServiceBus` 7.20.2 report `SequenceNumber = 0` and `EnqueuedTime = DateTimeOffset.MinValue`. Peeking the same queued messages reports distinct sequence numbers and valid enqueue timestamps.

This breaks applications that use received-message metadata for DLQ retention, inspection deduplication, or replay diagnostics. In PalCal, a fresh DLQ message appears thousands of years old and qualifies for immediate cleanup; subsequent messages with the same zero sequence number are treated as already inspected.

## Environment

- Source tested: upstream `main` at `efdedff5cf2956f28f2f7baede3a40ff0a7a301a`, on 2026-09-12.
- Runtime: packaged JVM server built with Java 25, with a real managed Artemis sidecar on Docker Desktop/Windows.
- Client: .NET 10, `Azure.Messaging.ServiceBus` 7.20.2.
- The Maven project reports 0.12.0, but this is a build of the newer source commit, not the published 0.12.0 image.

## Reproduction

Start floci-az with `FLOCI_AZ_SERVICES_SERVICE_BUS_MOCKED=false`. Create a namespace and an ordinary queue. Set `FLOCI_SB_CONNECTION_STRING` to its AMQP connection string and `FLOCI_SB_QUEUE` to the queue name, then run this console program with the SDK version above:

```csharp
using System.Text.Json;

using var timeout = new CancellationTokenSource(TimeSpan.FromSeconds(60));
var ct = timeout.Token;
var connectionString = Environment.GetEnvironmentVariable("FLOCI_SB_CONNECTION_STRING")!;
var queue = Environment.GetEnvironmentVariable("FLOCI_SB_QUEUE")!;

await using var client = new ServiceBusClient(connectionString);
await using var sender = client.CreateSender(queue);
await sender.SendMessagesAsync(new[]
{
    new ServiceBusMessage("one") { MessageId = "one" },
    new ServiceBusMessage("two") { MessageId = "two" }
}, ct);

await using var receiver = client.CreateReceiver(queue);
foreach (var message in await receiver.PeekMessagesAsync(2, cancellationToken: ct))
    Print("peek", message);

for (var i = 0; i < 2; i++)
{
    var message = await receiver.ReceiveMessageAsync(TimeSpan.FromSeconds(5), ct)
        ?? throw new Exception("Missing queue message");
    Print("receive", message);
    await receiver.DeadLetterMessageAsync(message, "audit-poison", "metadata probe", ct);
}

await using var dlq = client.CreateReceiver(queue,
    new ServiceBusReceiverOptions { SubQueue = SubQueue.DeadLetter });
foreach (var message in await dlq.PeekMessagesAsync(2, cancellationToken: ct))
    Print("dlq-peek", message);

for (var i = 0; i < 2; i++)
{
    var message = await dlq.ReceiveMessageAsync(TimeSpan.FromSeconds(5), ct)
        ?? throw new Exception("Missing DLQ message");
    Print("dlq-receive", message);
    await dlq.CompleteMessageAsync(message, ct);
}

static void Print(string operation, ServiceBusReceivedMessage message) =>
    Console.WriteLine(JsonSerializer.Serialize(new
    {
        operation,
        message.MessageId,
        message.SequenceNumber,
        message.EnqueuedTime,
        message.DeadLetterReason
    }));
```

## Actual behavior

Selected fields from the live run:

| Operation | MessageId | SequenceNumber | EnqueuedTime |
|---|---|---:|---|
| peek | one | 333 | 2026-09-12T11:51:23.544+00:00 |
| peek | two | 334 | 2026-09-12T11:51:23.545+00:00 |
| receive | one | 0 | 0001-01-01T00:00:00+00:00 |
| receive | two | 0 | 0001-01-01T00:00:00+00:00 |
| dlq-peek | one | 357 | 2026-09-12T11:51:23.544+00:00 |
| dlq-peek | two | 359 | 2026-09-12T11:51:23.545+00:00 |
| dlq-receive | one | 0 | 0001-01-01T00:00:00+00:00 |
| dlq-receive | two | 0 | 0001-01-01T00:00:00+00:00 |

Both DLQ receives retain `DeadLetterReason: "audit-poison"`, so dead-letter settlement and body delivery work; the failure is specific to the broker metadata.

A separate probe created a topic subscription with `RequiresSession=true`, sent a message with `SessionId="aggregate-1"`, accepted that session, received and dead-lettered the message, then received it from the subscription DLQ using an ordinary receiver. Both the session receive and the subscription DLQ receive returned the same zero sequence number and year-0001 timestamp.

## Expected behavior

Normal receives and DLQ receives must expose the broker metadata appropriate to the entity, consistently with peeking. They must not substitute a constant zero identifier or a minimum timestamp.

- [SequenceNumber](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.sequencenumber?view=azure-dotnet) is the unique broker-assigned number.
- [EnqueuedTime](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.enqueuedtime?view=azure-dotnet) is the UTC time at which the message was accepted and stored.

The different sequence values observed before and after dead-lettering are included as evidence, not as a separate claim about DLQ sequence preservation.

## Source evidence and application impact

[`ServiceBusPeekPlugin.encodePeekedMessage`](https://github.com/floci-io/floci-az/blob/efdedff5cf2956f28f2f7baede3a40ff0a7a301a/src/artemis-plugin/java/io/floci/az/artemis/ServiceBusPeekPlugin.java#L186) explicitly adds `x-opt-sequence-number` and `x-opt-enqueued-time` to the peek response. The normal delivery path does not supply equivalent annotations in the tested build, matching the receive/peek discrepancy.

PalCal maps the received SDK values directly into its DLQ model, then:

- [Deduplicates inspection by SequenceNumber](https://github.com/pal-cal/PalCal/blob/55f68e44f9ce36bfa63e4eca197bc4bc0bf2a6a1/src/PalCal.Messaging/Dlq/DlqDrainService.cs#L136).
- [Completes messages older than the residency cap](https://github.com/pal-cal/PalCal/blob/55f68e44f9ce36bfa63e4eca197bc4bc0bf2a6a1/src/PalCal.Messaging/Dlq/DlqDrainService.cs#L214).

The downstream cleanup impact is derived from those code paths; the full PalCal cleanup service was not executed in this probe.

## Suggested acceptance

- Add failing-first .NET SDK coverage for two distinct received messages having distinct valid sequence numbers and realistic enqueue timestamps.
- Cover ordinary queues, session-enabled subscriptions, and both DLQ receive paths.
- Compare peek and receive metadata for the same message in the same entity.
- Preserve the metadata through redelivery and dead-lettering according to Azure semantics, including streamed large-message delivery.
- Verify that newly dead-lettered messages do not appear past a retention cutoff and that distinct messages remain distinguishable to an inspection loop.

This is separate from open PR #298 (PeekLock expiry and lock metadata) and merged PR #232 (entity runtime metadata). Neither addresses these received-message fields.

thomhurst:
Service Bus receives currently return a zero SequenceNumber and minimum EnqueuedTime even when peeking the same message returns valid broker metadata. Add entity-local broker IDs and ingress timestamps to outgoing message annotations while preserving delivery-specific lock annotations and shared stored state.

Standard messages use a delivery copy. Large messages keep their streamed bodies and rebuild only their headers; large-message copies also preserve current ingress timestamps and dead-letter properties. Covers queues, subscriptions, sessions, dead-letter queues, ReceiveAndDelete, abandon, and lock-expiry redelivery.

Includes .NET SDK regressions using Azure.Messaging.ServiceBus 7.20.2, Java and Node.js compatibility assertions, and focused tests of metadata projection and large-message copying. Real encoded-message tests verify metadata and lock annotations without mutating stored messages. SDK assertions also preserve correlation IDs, content types, and custom properties across redelivery and dead-lettering. Integrated with the merged PeekLock implementation from #298.

Validation against current upstream main:
- Service Bus and Artemis patch unit/integration tests: 186 passed.
- Live .NET Service Bus SDK tests: 25 passed with a real Artemis 2.44.0 sidecar; all 10 metadata cases passed again with the expanded property assertions.
- Live Java Service Bus SDK tests: 18 passed initially; the gated session-expiry case also passed separately with SERVICEBUS_SESSION_LOCK_DURATION_SECONDS=60. All 19 cases exercised.
- Live Node.js Service Bus SDK tests: 5 passed.
- JVM package build and `git diff --check` passed.

Documented limitation: sequence numbers follow Artemis message IDs, as peeking already does. They may have gaps and may change when a message moves to a dead-letter queue; the original enqueue timestamp is preserved.

Fixes #302.

### DIFF
--- a/compatibility-tests/sdk-test-dotnet/SdkTestDotnet.csproj
+++ b/compatibility-tests/sdk-test-dotnet/SdkTestDotnet.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.4.0" />
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.20.1" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.20.2" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.11.0" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.29.2" />
     <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.62.0" />
--- a/compatibility-tests/sdk-test-dotnet/ServiceBusMetadataCompatibilityTests.cs
+++ b/compatibility-tests/sdk-test-dotnet/ServiceBusMetadataCompatibilityTests.cs
@@ -1,0 +1,172 @@
+using Azure.Messaging.ServiceBus;
+using Azure.Messaging.ServiceBus.Administration;
+
+namespace FlociAz.Compatibility;
+
+[NotInParallel]
+public sealed class ServiceBusMetadataCompatibilityTests
+{
+    [Test]
+    [Arguments(false, false, false)]
+    [Arguments(false, false, true)]
+    [Arguments(false, true, false)]
+    [Arguments(false, true, true)]
+    [Arguments(true, false, false)]
+    [Arguments(true, false, true)]
+    [Arguments(true, true, false)]
+    [Arguments(true, true, true)]
+    [Timeout(90_000)]
+    public async Task ReceiveMetadataMatchesPeekThroughRedeliveryAndDeadLetter(
+        bool subscription, bool session, bool large, CancellationToken cancellationToken)
+    {
+        var admin = new ServiceBusAdministrationClient(ConnectionString(administration: true));
+        await using var client = new ServiceBusClient(ConnectionString());
+        string entity = $"metadata-{Guid.NewGuid():N}";
+        if (subscription)
+        {
+            await admin.CreateTopicAsync(entity, cancellationToken);
+            await admin.CreateSubscriptionAsync(new CreateSubscriptionOptions(entity, "consumer")
+            {
+                RequiresSession = session
+            }, cancellationToken);
+        }
+        else
+        {
+            await admin.CreateQueueAsync(new CreateQueueOptions(entity) { RequiresSession = session }, cancellationToken);
+        }
+
+        try
+        {
+            DateTimeOffset beforeSend = DateTimeOffset.UtcNow.AddSeconds(-5);
+            string body = large ? new string('x', 200_000) : "metadata";
+            await using ServiceBusSender sender = client.CreateSender(entity);
+            for (int i = 0; i < 2; i++)
+            {
+                await sender.SendMessageAsync(new ServiceBusMessage(body + i)
+                {
+                    MessageId = i.ToString(), SessionId = session ? "session" : null
+                }, cancellationToken);
+            }
+            DateTimeOffset afterSend = DateTimeOffset.UtcNow.AddSeconds(5);
+            await using ServiceBusReceiver receiver = await CreateReceiver(client, entity, subscription, session, cancellationToken);
+            var peeked = (await receiver.PeekMessagesAsync(2, cancellationToken: cancellationToken))
+                .ToDictionary(message => message.MessageId);
+            await Assert.That(peeked.Count).IsEqualTo(2);
+
+            var received = new List<ServiceBusReceivedMessage>();
+            for (int i = 0; i < 2; i++)
+            {
+                ServiceBusReceivedMessage message = await Receive(receiver, cancellationToken);
+                await CheckMetadata(message, peeked[message.MessageId], beforeSend, afterSend);
+                await Assert.That(message.Body.ToString()).IsEqualTo(body + message.MessageId);
+                received.Add(message);
+            }
+            await Assert.That(received.Select(message => message.SequenceNumber).Distinct().Count()).IsEqualTo(2);
+            foreach (ServiceBusReceivedMessage message in received)
+            {
+                await receiver.AbandonMessageAsync(message, cancellationToken: cancellationToken);
+            }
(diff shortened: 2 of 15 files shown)


hectorvent:
Thank you, and this is a clean one to review. I checked both annotation names at the SDK source rather than taking them from the description: `x-opt-sequence-number` and `x-opt-enqueued-time` are exactly what `AmqpMessageConstant` declares and what `ReactorReceiver` reads off the delivery, so clients will pick these up without any mapping on their side.

Keeping the change in the Artemis patch tree beside the PeekLock work is the right layer for broker assigned metadata, and preserving the lock annotations rather than rebuilding the whole delivery is what makes it compose with that.

The documented limitation is the part I would have asked for: sequence numbers following Artemis message IDs means gaps and a possible change on dead-lettering, which real Service Bus does not have. Good that it is written down on the service page rather than left implicit.

No blockers from my side.

hectorvent (approved):

PR #308, opened 2026-09-13
feat(servicebus): support message and session lock renewal

thomhurst:
Adds Azure AMQP message and session lock renewal. A receiver can extend its lock, process past the original deadline, and complete successfully. Expired or settled locks return the SDK's lock-lost error instead of timing out.

Renewal replaces the expiry deadline on the owning connection event loop. Old timers cannot expire a renewed lock. Delivery tags receive fresh token values even when their buffers are reused, and management requests replenish receiver credits.

Validation:
- Regression tests failed before implementation: missing renewal method in Java, and real .NET SDK renewal requests timed out against unchanged main.
- 191 focused Java Service Bus/AMQP tests passed.
- All 28 existing and new .NET Service Bus tests passed; the final renewal slice passed all four tests, including 1,050 consecutive renewals on one link.
- Full Java suite: 1,229 tests, one failure, 14 skipped. The sole failure is the unchanged `SqlProvisioningServiceTest.boundsRetainedTerminalOperations` timestamp-tie assertion. Reproduced on unchanged main by repeatedly invoking that existing test; it is unrelated to this PR.

Deferred message locks and session state remain unsupported. Protocol reference: https://learn.microsoft.com/azure/service-bus-messaging/service-bus-amqp-request-response


hectorvent:
Thank you, and this one is well aimed at the wire. I checked all four vendor symbols and both reply bodies at the SDK source rather than from the description, and they match, including the part that usually goes wrong: the message path has to return `expirations` as a `Date[]`, because [the serializer gates on `instanceof Date[]`](https://github.com/Azure/azure-sdk-for-java/blob/e2eddb6460286bb37546d75b0aeac0eb78a24d89/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java#L335) and quietly returns an empty list for a `List<Date>`.

**(follow-up, separate PR)** The bot's session timer point is right, and your own message lock code is the fix: `ServiceBusMessageLockSupport` keeps the `ScheduledFuture` and cancels it before rescheduling, while `ServiceBusSessionSupport.scheduleExpiration` drops the handle. Correctness is safe either way, since the two argument `SESSION_OWNERS.remove(session, owner)` stops a stale timer expiring a renewed lock, so this is only retention until the old deadline passes.

**(follow-up, separate PR)** Renewal is exercised by the .NET suite only, and that `Date[]` detail is Java specific, so a Java `renewMessageLock` case would pin it.

No blockers from my side.

hectorvent (approved):

PR #309, opened 2026-09-14
feat(keyvault): support self-signed certificate lifecycle

thomhurst:
Adds self-signed Key Vault certificate issuance and lifecycle support. Applications can create an RSA or EC certificate through `CertificateClient`, retrieve its matching private key through `SecretClient`, and use the exported PFX for signing. Certificate, key, and secret versions share the same version identifier.

Supports certificate/version listing, policy reads, completed-operation polling, PKCS12/PEM export, non-exportable certificates, and delete/recover/purge of the certificate with its associated objects. Uses the existing Key Vault storage backend and cryptography implementation.

Validation:
- The new regression test returned 404 before implementation; the .NET SDK also failed against unchanged main.
- All 82 focused Key Vault tests passed after review fixes, including ownership conflicts, reused-key parameters, and rotation-policy recovery.
- All 37 Java Key Vault SDK tests passed, including the new certificate lifecycle and PFX test.
- All six .NET Key Vault tests passed. The Certificates 4.9.1 SDK test covers: create/poll, PFX load, RSA sign/verify, multiple versions, delete, recover, and purge.
- Focused tests cover EC certificates, non-exportable secrets, SANs, unsupported issuers, invalid key sizes, and malformed requests.

Scope and limitations are documented in `docs/services/key-vault.md`. External issuers, import/merge, policy/property updates, scheduled renewal, UPN SANs, and certificate backup/restore remain unsupported. Associated deleted keys and secrets are recovered with the certificate rather than exposed through independent deleted-object APIs.


hectorvent:
Thank you, this is a big surface landed carefully, and the scope note listing what is deliberately unsupported makes it much easier to review. The status codes match the 7.4 Certificates spec where it matters: create is 202, purge 204, recover 200, and the `CertificateOperation` body uses the spec's own field names with `issuer` as `Self`.

**(blocking)** The bot's deletion point holds up, and the root cause looks like one guard rather than the delete loop. Deleting certificate `foo` tombstones everything under `keys/foo` and `secrets/foo`, which is correct for objects the certificate owns. The gap is that creation only refuses when a *deleted* key or secret of that name exists, never a live one, so a standalone key created first is silently swept later. Real Key Vault shares that namespace and rejects the collision up front, so extending the existing 409 check to live objects would match Azure and close this.

**(follow-up, separate PR)** This adds a third private Key Vault error builder alongside the two already in the package. A shared helper, the way `AcrErrors` works, would be tidier.

The deletion guard is the only thing blocking from my side.

hectorvent:
Thank you for the quick turnaround. The guard now refuses a live key or secret of the same name on both create and delete, and the new test drives exactly the sequence that was reachable before, so the blocking point is closed. Moving issuance, delete, recover and purge onto batched writes at the same time is a welcome improvement.

Rotation policies on certificate owned keys surviving renewal and recovery reads right, and the reuse_key size and curve checks close the second bot finding cleanly.

The shared error helper stays a follow-up for a separate PR. No blockers from my side.

hectorvent (approved):

PR #310, opened 2026-09-14
fix(servicebus): cancel superseded session expiry tasks

thomhurst:
Cancels session expiration tasks when their owner is replaced by renewal, released, or expired. The existing owner comparison still rejects callbacks that were already queued before cancellation.

Follow-up requested in #308, which is now merged. Also adds Java SDK renewal coverage: `renewMessageLock` must deserialize the expiration timestamp array, extend the deadline, and allow completion after the original deadline.

Validation:
- The session release regression failed before the fix because its scheduled task was never cancelled.
- All 192 focused Service Bus/AMQP Java tests passed.
- All four .NET renewal compatibility tests passed, including session renewal and 1,050 consecutive message renewals.
- The new Java SDK renewal test passed against the real Artemis sidecar.


hectorvent:
Thank you for following up on #308 so quickly, and for bringing the Java SDK renewal test with it, which is the coverage that suite was missing.

The timer handling reads correctly on every path: release and expire cancel through the same compare and remove guard, renew cancels the superseded owner only after the replace succeeds, and a callback that was already queued still fails the owner check, exactly as the description says. Keeping the future on the owner record rather than in a side map is the right shape, since the record is what the map compares.

One note for confidence rather than a change: the unit test drives the real patched class through the jar loader and verifies the cancel on the mocked future, so it exercises the code that ships, not a copy.

No blockers from my side.

hectorvent (approved):

PR #311, opened 2026-09-14
feat(signalr): support Azure SignalR Default mode for ASP.NET Core hubs

thomhurst:
Adds Azure SignalR Default mode for ASP.NET Core hubs. Applications using `AddAzureSignalR` can negotiate clients through the emulator, invoke hub methods, and send messages across application server instances. JSON and MessagePack clients can share a hub.

Implements the Azure MessagePack service protocol over WebSockets: connection lifecycle, keep-alives, connection/user/group/broadcast delivery, acknowledged group membership, and preferred/required server stickiness. Signed client and server tokens are checked for signature, audience, and lifetime. Live connections and memberships are transient; no sidecar is required.

Validation:
- The real SDK regression failed against unchanged main because the service endpoint was missing.
- Full Java suite: 1,233 tests, zero failures, 14 skipped.
- 18 focused protocol, authentication, routing, and routing-table tests passed.
- .NET SDK compatibility passed with Microsoft.Azure.SignalR 1.33.1 and SignalR clients 10.0.11: two application servers, an isolated third application, JSON/MessagePack hub calls, groups, users, broadcasts, preferred stickiness, reconnect, and 35 seconds of idle time to verify keep-alives.

The supported surface and setup are documented in `docs/services/signalr.md`. This PR does not implement SSE/long polling, serverless upstreams, REST management, ARM provisioning, Entra authentication, migration, stateful reconnect, or server-to-client invocations that return results. Unsupported service message types close the server connection explicitly.

Protocol sources:
- https://github.com/Azure/azure-signalr/blob/dev/specs/ServiceProtocol.md
- https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs


hectorvent:
Thank you, this is a large new surface and it lands with the right evidence: a real ASP.NET Core host under Microsoft.Azure.SignalR 1.33.1 in the .NET suite, and CI ran that suite against the native image, so the new msgpack dependency is already proven there.

I checked the service protocol against the SDK source where the spec page is thin, and it matches: an empty handshake error string counts as success, OpenConnection is the [six slot form](https://github.com/Azure/azure-signalr/blob/dab640352d551798bd855157f09474a4b0facf5a/src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs#L420) with the hub protocol in extension key 3, and Ack status 1 and 2 are what the SDK reads.

The token check matches too: the audience [drops the port](https://github.com/Azure/azure-signalr/blob/dab640352d551798bd855157f09474a4b0facf5a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs#L240) and the signature uses the UTF-8 bytes of the key string, not its Base64 decoding.

**(follow-up, separate PR)** The SDK can also sign with HS512 when `AccessTokenAlgorithm` is set, which the verifier rejects today. The default is HS256, so nothing breaks out of the box.

No blockers from my side.

hectorvent (approved):

PR #312, opened 2026-09-14
docs(blob): document snapshot support

snazy:
## Summary

- Document modeled Blob snapshot creation, reads, deletion, and immutable snapshot behavior.
- Clarify snapshot-scoped user delegation SAS support (`sr=bs`).
- Retain Blob versioning and tiering as unsupported.

## Type of change

- [x] Docs / chore

## Azure Compatibility

Documents the supported Blob snapshot behavior and snapshot-scoped user delegation SAS authorization.

## Checklist

- [ ] New or updated integration test added


hectorvent:
Thank you for following up on the snapshot work with the docs. I checked each sentence against the handler and it all matches, including the `SnapshotsPresent` behaviour on both paths.

**(follow-up, separate PR)** Listing snapshots (`include=snapshots` on List Blobs) is not modeled yet. It might be worth one clause in the snapshot bullet so nobody expects them to show up in a listing, but that can come any time.

No blockers from my side.

hectorvent (approved):

PR #314, opened 2026-09-15
fix(servicebus): restore zero-based AMQP delivery counts

### LINKED ISSUE #313: [BUG] Service Bus: first session receive reports DeliveryCount 2 instead of 1 in 0.13.0
## Service

Azure Service Bus, using the managed Artemis sidecar and the .NET SDK.

## Azure API action

`ServiceBusClient.AcceptSessionAsync` followed by `ServiceBusSessionReceiver.ReceiveMessageAsync` on a session-enabled topic subscription. The incorrect count is also visible after explicit dead-lettering.

## Expected behavior

The first receive of a fresh message exposes `ServiceBusReceivedMessage.DeliveryCount == 1`. Neither a non-destructive peek nor session-lock renewal constitutes another delivery.

The [Azure .NET SDK contract](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.deliverycount?view=azure-dotnet) specifies an initial value of 1, incrementing after an explicit abandon or message-lock expiry.

## Actual behavior

With released `floci/floci-az:0.13.0`, the first receive reports **2**. This reproduces with and without a preceding peek:

```text
failed ServiceBus_SessionsFilterSubjectsDeduplicateAndSupportDeadLetters(True)
Expected to be 1, because the first receive is the first delivery, with or without a non-destructive peek
but found 2

failed ServiceBus_SessionsFilterSubjectsDeduplicateAndSupportDeadLetters(False)
Expected to be 1, because the first receive is the first delivery, with or without a non-destructive peek
but found 2
```

The direct SDK cases finish in approximately 545 ms each. They use a fresh session/message ID and do not abandon the message or wait for lock expiry before checking the count. No Wolverine consumer is running in this fixture.

An existing independent application test using Wolverine 6.16.0 also fails after a malformed payload is immediately dead-lettered without a broker retry:

```text
failed LiveConsumer_PoisonAndExhaustedTransientMessagesReachBrokerDlqWithoutBlockingSession
Expected to be 1, because malformed wire payloads are non-retriable poison messages
but found 2
```

These are three manifestations of the same observed count mismatch, not three separate defect reports. A previous diagnostic run checked that explicit DLQ transfer preserves the received count; that assertion passed. The count is already incorrect before dead-lettering, so the evidence does not establish a separate DLQ-transfer increment bug.

## Reproduction

The executed regression test is [FlociAzureCompatibilityTests.cs](https://github.com/pal-cal/PalCal/blob/ed2fadfb529f9088ff67c8418e5444d36a580e2b/tests/PalCal.Notifications.IntegrationTests/FlociAzureCompatibilityTests.cs#L115), introduced in [pal-cal/PalCal#8719](https://github.com/pal-cal/PalCal/pull/8719).

The relevant setup is a topic with duplicate detection enabled and a subscription with `RequiresSession=true`, a subject correlation rule, and a normal lock duration. The fixture uses real AMQP through Floci's owned Artemis sidecar. Run the test for both `peekBeforeReceive` arguments:

```powershell
dotnet test --project tests/PalCal.Notifications.IntegrationTests --treenode-filter '/**/FlociAzureCompatibilityTests/ServiceBus_SessionsFilterSubjectsDeduplicateAndSupportDeadLetters'
```

For a standalone reduction, use a dedicated topic/subscription with the same settings and a correlation rule whose subject is `delivery-count-probe`. Set `FLOCI_SB_CONNECTION_STRING`, `FLOCI_SB_TOPIC`, and `FLOCI_SB_SUBSCRIPTION`, then run the following with `Azure.Messaging.ServiceBus` 7.20.2. This reduction mirrors the executed test's operations; the linked PalCal test is the reproduction actually run.

```csharp

using var timeout = new CancellationTokenSource(TimeSpan.FromSeconds(45));
var token = timeout.Token;
await using var client = new ServiceBusClient(
    Environment.GetEnvironmentVariable("FLOCI_SB_CONNECTION_STRING")!);
var topic = Environment.GetEnvironmentVariable("FLOCI_SB_TOPIC")!;
var subscription = Environment.GetEnvironmentVariable("FLOCI_SB_SUBSCRIPTION")!;
await using var sender = client.CreateSender(topic);

foreach (var peekBeforeReceive in new[] { false, true })
{
    var session = Guid.NewGuid().ToString("N");
    var message = new ServiceBusMessage("payload")
    {
        MessageId = session,
        SessionId = session,
        Subject = "delivery-count-probe"
    };
    await sender.SendMessageAsync(message, token);
    await sender.SendMessageAsync(message, token); // Same ID: topic deduplication.
    await sender.SendMessageAsync(new ServiceBusMessage("wrong subject")
    {
        MessageId = Guid.NewGuid().ToString("N"),
        SessionId = session,
        Subject = "not-a-logical-event"
    }, token);

    await using var receiver = await client.AcceptSessionAsync(
        topic, subscription, session, cancellationToken: token);
    await receiver.RenewSessionLockAsync(token);
    if (peekBeforeReceive)
    {
        _ = await receiver.PeekMessageAsync(cancellationToken: token)
            ?? throw new Exception("Missing peeked message");
    }

    var received = await receiver.ReceiveMessageAsync(TimeSpan.FromSeconds(5), token)
        ?? throw new Exception("Missing received message");
    Console.WriteLine($"peek={peekBeforeReceive}, first DeliveryCount={received.DeliveryCount}, expected=1");
    await receiver.CompleteMessageAsync(received, token);
}
```

The duplicate send and session renewal are retained because they were present in the executed SDK test; neither has yet been independently excluded as a trigger. No claim is made here about ordinary non-session queues or large-message delivery.

## Environment

- Image: `floci/floci-az:0.13.0`, tested on 2026-09-15.
- Docker Desktop on Windows, with a real Floci-managed Artemis sidecar.
- .NET 10; `Azure.Messaging.ServiceBus` 7.20.2.
- Aspire 13.5.3; `CommunityToolkit.Aspire.Hosting.Floci` 13.5.1-beta.757.
- Per-instance Docker namespace and dynamically allocated AMQP ports.
- The reproducing fixture has no application consumers; this is not concurrent consumption by the full application fleet.

## Impact and suggested acceptance

This blocks PalCal's emulator migration: a non-retriable poison message appears to have been delivered twice. The adapter does not adjust or mask SDK metadata.

- Add failing-first SDK coverage for first delivery count 1 on session subscriptions, with and without peek/session renewal.
- Check that one explicit abandon produces the next expected count, rather than an offset sequence.
- Cover explicit dead-lettering without introducing a synthetic extra delivery.
- Verify the mapping between Artemis delivery state, the AMQP header, and the SDK's count. The exact source-level cause has not yet been established.

Related but distinct: #302 addressed missing sequence/enqueue metadata; this report concerns `DeliveryCount` on the released 0.13.0 image.

thomhurst:
## Summary

Fix first receives reporting `DeliveryCount = 2` in the .NET SDK. Remove the Artemis source replacement that changed `deliveryCount - 1` into `deliveryCount`, restoring a zero-based AMQP header. Retain the first-delivery header emission: omitting that header causes null/undefined metadata in the SDKs.

The wire count is `0` for first delivery and `1` after one abandon. .NET adds one when exposing `DeliveryCount`; Java and JavaScript expose the raw header. Correct the Java compatibility test's initial expectation and replace its permissive redelivery bound with an exact assertion. Document these SDK differences.

For streamed messages, move count normalization from message-lock handling to shared Service Bus metadata handling. Session receivers and ReceiveAndDelete have no individual message deadline, but still require an explicit delivery-count header. Preserve other header fields and the stored message; leave non-Service-Bus deliveries unchanged.

Closes #313

## Regression coverage

- .NET: initial and redelivery counts across queues/subscriptions, sessions, and standard/large messages; ReceiveAndDelete; immediate session DLQ transfer with and without a preceding peek and session-lock renewal.
- Java: exact raw counts before and after abandon.
- JavaScript: exact subscription receive/redelivery counts after a non-destructive peek.
- Packaged Artemis helper tests: streamed counts without individual message locks, header preservation, and non-Service-Bus behavior.

Before the fix, 10 of the 12 .NET metadata cases failed, and the strengthened Java and JavaScript cases also failed against unchanged 0.13.0 source. No SDK-side offset or per-client behavior is introduced.

SDK reference: [.NET DeliveryCount](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusreceivedmessage.deliverycount?view=azure-dotnet), [Java deserialization](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/servicebus/azure-messaging-servicebus/src/main/java/com/azure/messaging/servicebus/ServiceBusMessageSerializer.java).

## Validation

- `./mvnw.cmd package -Dtest=ServiceBus*,ArtemisPatchVersionsTest`: 192 passed.
- .NET Service Bus suites: 31 passed, zero skipped.
- Java SDK Service Bus and rule suites: 26 passed, 1 existing skip requiring the optional configured session-lock duration.
- Node Service Bus suite: 5 passed.
- Full Java suite: 1,302 passed, 1 pre-existing SQL-retention failure, 19 existing skips.

SDK runs use .NET 10 / Azure.Messaging.ServiceBus 7.20.2, Java SDK 7.17.4, and JavaScript SDK 7.9.5 against the packaged JVM server and a real owned Artemis sidecar on Docker Desktop. Self-reviewed the full diff and applied code-simplifier review.

One concurrent cross-SDK run stalled in a Java send and was interrupted after capturing its thread dump. The final Java verification uses a separate broker; mixed-SDK concurrency has not been established by these checks.

## Existing full-suite failure

The full Maven run had 1 failure in `SqlProvisioningServiceTest.boundsRetainedTerminalOperations` (1,322 tests, 19 existing skips). Repeating that unchanged test against baseline production source at `18a139b` reproduced 43 failures in 50 invocations. SQL retention sorts a `ConcurrentHashMap` snapshot only by `endTime`; equal timestamps leave eviction order undefined, while the test assumes the first inserted operation is always removed. No SQL code or test changes are included here.


hectorvent:
Thank you for tracking this down to the Artemis patch. The before and after counts across .NET, Java and JavaScript make the fix easy to trust.

The wire contract checks out: the .NET SDK adds one to the AMQP header on receive but not on peek ([AmqpMessageConverter.cs#L143](https://github.com/Azure/azure-sdk-for-net/blob/e345636942bb5b24c090a6df28e6b4bf65f8f8e4/sdk/servicebus/Azure.Messaging.ServiceBus/src/Amqp/AmqpMessageConverter.cs#L143)), while Java and JavaScript surface the raw header, so zero on the first delivery is what all three expect. Swapping the old `>= 2` bound for exact assertions is what should have caught this in the first place.

The `SqlProvisioningServiceTest` failure you describe is a known flaky test on our side, and your diagnosis of the tied `endTime` ordering is helpful. Nothing to do here.

No blockers from my side.

hectorvent (approved):

PR #315, opened 2026-09-18
feat(blob): add Copy Blob and Copy Blob From URL

kozhinhikkodan-dev:
## Summary

Adds support for the two Copy Blob REST API operations that were previously returning `501 NotImplemented`:

- **Copy Blob** (async) — `PUT /{container}/{blob}` with `x-ms-copy-source` header, returns `202 Accepted`
- **Copy Blob From URL** (sync) — same PUT with `x-ms-requires-sync: true`, returns `200 OK`

Both operations complete synchronously within the emulator (no async polling needed for local dev), matching how Azurite handles this.

The `isPutBlob()` guard already excluded `x-ms-copy-source` requests from landing in `putBlob()` — they were falling through to the `notImplemented()` catch-all. This PR adds the dispatch branch and handler method for them.

**Supported:**
- Same-container, cross-container, and cross-account copies (within the same emulator instance)
- Absolute URLs and relative paths as the copy source
- Azure metadata semantics (inherit from source, or replace when `x-ms-meta-*` headers are provided)
- Standard response headers (`x-ms-copy-id`, `x-ms-copy-status`, `ETag`, `Last-Modified`)
- Conditional header validation and lease checks on the destination blob

**Intentionally out of scope (follow-up PRs):**
- Cross-external-URL copies, conditional source headers, append/page blob copy, copy abort, async status polling

## Type of change

- [ ] Bug fix (`fix:`)
- [x] New feature (`feat:`)
- [ ] Docs / chore

## Azure Compatibility

Verified against the [Azure Storage Blob REST API spec](https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob) for Copy Blob and [Put Blob From URL](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url). Wire protocol validated with RestAssured integration tests matching the Azure SDK request shapes — 10 new tests in `BlobServiceTest.java` covering async/sync copy, metadata inheritance/replacement, cross-container, overwrite, relative paths, and error cases (404, 400).

## Checklist


hectorvent:
Thank you for picking up Copy Blob. The metadata inherit and replace cases are exactly the semantics SDK users trip over, and they are well covered.

**(blocking)** CI is red on `BlobCompDispatchTest.copyBlobIsNotMistakenForPutBlob`, which still expects 501 (greptile spotted this too). Could it assert the new copy behaviour while keeping its "no empty blob" guarantee?

**(blocking)** Copy Blob From URL answers 202, not 200 ([blob.json#L6195](https://github.com/Azure/azure-rest-api-specs/blob/7efab4ca79331ad2ac48eb67fb944d030745918d/specification/storage/data-plane/Microsoft.BlobStorage/stable/2021-12-02/blob.json#L6195)), and the Java SDK treats any other status as an error, so `copyBlobFromUrlSyncReturns200` needs flipping. Put Blob From URL is a separate operation (`x-ms-blob-type: BlockBlob`, 201, content headers) that this branch now catches; excluding it keeps it at 501. A Java SDK test would pin both.

**(blocking)** Dropping the host and query means `?snapshot=` silently copies the live blob, and host style URLs are rejected. Parsing with `java.net.URI`, as [Azurite does](https://github.com/Azure/Azurite/blob/8b4e9258cc7d9f1ea46fe8dfeb534cc24bd18f16/src/blob/handlers/BlobHandler.ts#L646), fixes both.

Those three unblock it.

PR #317, opened 2026-09-20
docs: correct the routing, logging and compat-suite guidance

hectorvent:
## Summary

Corrects guidance in `AGENTS.md` and the two service tables that had drifted from the code.

`AGENTS.md` still described the pre-`ServiceRoutes` design: it told an agent adding a service to add suffix detection in `AzureRoutingFilter` and a `case` to `AzureServiceRegistry.isEnabled()`, and it never mentioned `routes()` or `ServiceRoutes` at all. Routing tables are assembled at `@PostConstruct` from each handler's `routes()`, and `isEnabled()` delegates to `handler.enabled(serviceType)`, so both steps now describe that, name the `RoutingTableAssemblyTest` golden, and note `LITERAL_ROUTE_SERVICE_TYPES` as the one remaining per-service edit.

Three smaller corrections in the same file: the Logging rule is scoped to `src/main/java`, since `src/artemis-*/` compiles into the Artemis sidecar where slf4j is the facade on the broker's classpath; the Makefile/CI env sync table is split into its suite-container and emulator-container axes, with four settings both sides pass on the emulator container that were unrecorded; and `sdk-test-dotnet` is added to the suite inventory.

`README.md` gains the same `sdk-test-dotnet` row (it runs in the CI matrix and is the only suite covering SignalR), loses the per-suite test count column, which was stale in six of its seven rows, and has four coverage cells corrected against what the suites now contain. The SignalR rows in `README.md` and `docs/services/index.md` named neither of the routes that serve it; they now list `/server/`, `/client/` and `*.service.signalr.net`, and the docs row gains the status marker every other service row carries.

The `{account}-signalr` suffix is deliberately not listed: the audience it builds carries the account segment, which the SDK never produces, so a real SDK token is rejected on that route.

## Type of change

- [ ] Bug fix (`fix:`)
- [ ] New feature (`feat:`)
- [x] Docs / chore

## Azure Compatibility

N/A, documentation only, no emulated surface changes. Every claim was verified against the code at head: the routing assembly in `AzureRoutingFilter`, `AzureServiceRegistry.isEnabled()`, the patch-tree build in `pom.xml`, each row of the env table against `Makefile:40-57` and `.github/workflows/compatibility.yml:94-120`, the suite contents under `compatibility-tests/`, and the SignalR audience construction in `SignalRWebSockets.java:67`.

## Checklist

- [ ] `./mvnw test` passes locally, not run: no Java changed. `mkdocs build --strict` passes at this head.
- [ ] New or updated integration test added, none: documentation only.


hectorvent:
Thank you, three of the four were real and are fixed in b415243.

The logging rule now names the facade per tree: slf4j in the AMQP patch tree, `System.Logger` in the
plugin tree, nothing in the proton patch tree. The routing step now separates table-routed services,
which need no filter edit, from stage-routed ones, which need a stage added and ordered, and says
plainly that `LITERAL_ROUTE_SERVICE_TYPES` is a declaration for the drift test rather than a dispatch
mechanism. The env-sync section now documents the `compat-docker` exception, where one shared emulator
carries the Java lock duration for every suite.

On the SignalR route, the omission was deliberate and I should have said so on the page rather than
leaving it silent. The `-signalr` suffix is registered, but the SDK builds the token audience from the
endpoint's scheme and host with no path segment
([ServiceEndpoint.cs#L242](https://github.com/Azure/azure-signalr/blob/dab640352d551798bd855157f09474a4b0facf5a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs#L242)),
and appends only `client` or `server` to it
([DefaultServiceEndpointGenerator.cs#L63](https://github.com/Azure/azure-signalr/blob/dab640352d551798bd855157f09474a4b0facf5a/src/Microsoft.Azure.SignalR/EndpointProvider/DefaultServiceEndpointGenerator.cs#L63)),
so a request through the prefixed path carries an audience that cannot match and gets a 401. Rather
than list it as a routing option, the service page now explains that it exists and why it cannot carry
an SDK connection.

